PR #2, opened 2022-03-09
Verify parachain headers

Wizdave97:
To be rebased after the first PR is merged


src/lib.rs
```diff
@@ -27,6 +27,7 @@ use crate::primitives::{
 };
 use crate::traits::{AuthoritySet, MmrState, StorageRead, StorageWrite};
 pub use beefy_primitives;
+pub use sp_core;
```
seunlanlege, line 30: any reason why?

Wizdave97: I didn't want to add  them as dependencies in ibc-rs.

seunlanlege: nahh, i think its fine

PR #11, opened 2022-05-11
Refactor to use host functions for all hashing and signature verifications

Wizdave97:


Wizdave97:
CI seems not to be working

PR #36, opened 2022-09-23
Update dependencies and fix compilation errors

vmarkushin:


algorithms/beefy/prover/src/lib.rs
```diff
@@ -307,7 +307,7 @@ where
 			signed_commitment.commitment.block_number.into();
 		let block_hash = self.relay_client.rpc().block_hash(Some(subxt_block_number)).await?;
 
-		let current_authorities = api.storage().beefy().authorities(block_hash).await?.0;
+		let current_authorities = api.storage().beefy().authorities(block_hash).await?;
```
seunlanlege, line 310: This isn't compatible with polkadot v0.9.27

light-clients/ics07-tendermint/src/client_state.rs
```diff
@@ -143,7 +143,7 @@ impl<H> ClientState<H> {
 	}
 
 	pub fn client_type() -> ClientType {
-		"07-tendermint".to_string()
+		"07-tendermint".to_owned()
```
seunlanlege, line 146: alloc::borrow::ToOwned is not in scope. Try compiling for wasm target

seunlanlege (changes requested):
Seems to have reverted updates for polkadot-v0.9.27 and not checked for wasm target.

algorithms/beefy/prover/src/lib.rs
vmarkushin: It's strange, because this code wasn't compiling

PR #84, opened 2022-10-31
Fishing mode

vmarkushin:
Closes: #75


algorithms/grandpa/prover/Cargo.toml
```diff
@@ -31,6 +31,7 @@ sp-finality-grandpa = { git = "https://github.com/paritytech/substrate", branch
 beefy-prover = { path = "../../beefy/prover" }
 primitives = { package = "grandpa-light-client-primitives", path = "../primitives" }
 light-client-common = { path = "../../../light-clients/common" }
+ics10-grandpa = { path = "../../../light-clients/ics10-grandpa" }
```
seunlanlege, line 34: Prover shouldn't depend on verifier/light client. Use polkadot primitives directly

contracts/pallet-ibc/src/impls.rs
```diff
@@ -66,30 +66,27 @@ where
...
+					},
 				}
-				(events, logs, errors)
-			},
-		);
+				(events, logs)
+			});
 
 		log::trace!(target: "pallet_ibc", "logs: {:#?}", logs);
-		log::trace!(target: "pallet_ibc", "errors: {:#?}", errors);
```
seunlanlege, line 84: Still log errors

vmarkushin: What do you mean?

contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -75,10 +84,76 @@ impl tendermint_light_client_verifier::host_functions::HostFunctionsProvider
 
 impl ics07_tendermint::HostFunctionsProvider for HostFunctionsManager {}
 
+pub struct GrandpaHederHashesStorageInstance;
```
Wizdave97, line 87:
```suggestion
pub struct GrandpaHeaderHashesStorageInstance;
```

contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -75,10 +84,76 @@ impl tendermint_light_client_verifier::host_functions::HostFunctionsProvider
...
 	fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool {
 		pub_key.verify(&msg, sig)
 	}
+
+	fn insert_relay_header_hashes(new_hashes: &[<Self::Header as Header>::Hash]) {
+		if new_hashes.is_empty() {
+			return
+		}
+
+		GrandpaHeaderHashesSetStorage::mutate(|hashes_set| {
```
Wizdave97, line 130: why can't we use only one of  `GrandpaHeaderHashesSetStorage` or `GrandpaHeaderHashesStorage`, they are storing the same values

hyperspace/primitives/src/lib.rs
```diff
@@ -366,6 +384,18 @@ pub trait Chain: IbcProvider + KeyProvider + Send + Sync {
...
 	) -> Result<(sp_core::H256, Option<sp_core::H256>), Self::Error>;
+
+	/// Returns an [`AnyClientMessage`] that triggered an [`IbcEvent`] at index `event_index`
+	/// on this chain.
+	///
+	/// Parameters `host_block_hash` and `transaction_index` identify the transaction that triggered
+	/// the event.
+	async fn query_client_message(
+		&self,
+		host_block_hash: [u8; 32],
```
Wizdave97, line 395: I think we can encapsulate these parameters in an enum with a variant for Parachains, because these current parameters seem parachain specific, when other chains are supported, the enum can be modified with variants for those new chains.

hyperspace/parachain/src/chain.rs
```diff
@@ -180,4 +198,163 @@ where
...
+			AnyClientMessage::Grandpa(ClientMessage::Header(header)) => {
+				let target_block_number = header
+					.finality_proof
+					.unknown_headers
+					.iter()
+					.max_by_key(|h| h.number)
+					.expect("unknown_headers always contain at least one header; qed")
+					.number;
+				let finalized_block_number =
+					*self.relay_client.rpc().block(None).await?.unwrap().block.header.number();
```
Wizdave97, line 295:
```suggestion
					*self.relay_client.rpc().block(None).await?.unwrap().block.header.number();
```
We should handle this error instead of unwrapping.

hyperspace/parachain/src/chain.rs
```diff
@@ -180,4 +198,163 @@ where
...
+				// We require a proof for the block number that may not exist on the relay chain.
+				// So, if it's greater than the latest block block the relay chain, we use the
```
Wizdave97, line 297:
```suggestion
				// So, if it's greater than the latest block block on the relay chain, we use the
```

contracts/pallet-ibc/src/light_clients.rs
vmarkushin: Because we need to keep order of the headers (to remove the oldest ones)

hyperspace/primitives/src/lib.rs
vmarkushin: Yes, this will be changed, Seun will suggest some solution

contracts/pallet-ibc/src/light_clients.rs
Wizdave97: But BoundedVec would always keep the items in the order of insertion.

hyperspace/parachain/src/chain.rs
vmarkushin: Isn't there always a block available?

contracts/pallet-ibc/src/light_clients.rs
vmarkushin: Yes, but we can't access an element in `BoundedVec` in O(1) by the hash

hyperspace/parachain/src/chain.rs
Wizdave97: Ideally there should be.

contracts/pallet-ibc/src/light_clients.rs
Wizdave97: Yeah that's true.

hyperspace/parachain/src/chain.rs
```diff
@@ -203,17 +205,108 @@ where
...
+				.await?;
+			match maybe_hash {
+				Some(hash) => break hash,
+				None => {
+					if now.elapsed() > Duration::from_secs(20) {
+						return Err(Error::from("Timeout while waiting for block".to_owned()))
+					}
+					sleep(Duration::from_millis(100)).await;
+				},
+			}
```
seunlanlege, line 231: hmmmm

vmarkushin: I had this problem of missing blocks at the beginning of implementation. But maybe I was doing something wrong and now there is no such problem.

seunlanlege:
This is blocked on CI

hyperspace/parachain/src/chain.rs
Wizdave97: I don't understand why you were having that issue, the event subscription api is using block header subscription internally, before fetching the events from storage, so by the time this function is called the block should be available, you can try it again without this hack and see if it works.

Wizdave97 (commented):
implementation looks solid to me, save the hack for fetching block hash.

hyperspace/testsuite/Cargo.toml
```diff
@@ -21,17 +21,27 @@ ibc-proto = { path = "../../ibc/proto" }
 tendermint-proto = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8" }
 
 hyperspace-core = { path = "../core", features = ["testing"] }
+hyperspace-parachain = { path = "../parachain", features = ["testing"] }
 hyperspace-primitives = { path = "../primitives", features = ["testing"] }
```
seunlanlege, line 25: Let's use the `build-metadata-from-ws` feature for tests.

contracts/pallet-ibc/src/impls.rs
seunlanlege: Don't remove the error logging, it's still useful

hyperspace/testsuite/Cargo.toml
vmarkushin: The CI fails because of this due to a connection error...

seunlanlege:
@vmarkushin please update this branch with master so we can run integration tests on this before we merge

seunlanlege:
ok we've merged the proper ci checks, @vmarkushin please pull so we can see the integration tests run on this pr

blasrodri:
I'm trying to fix this. Currently experiencing some issues w/ the generated code and some _apparently_ breaking changes. Investigating.

blasrodri:
@Wizdave97 we've been trying to fix this for some time with @vmarkushin and unfortunately we are failing to do so.
So far we can see that the `Event` type defined on `pallet-ibc` does not match the one generated in docker through `subxt`. 
We've tried cleaning docker cache but that did not help the problem.
Can you take a look at this, to understand what are the discrepancies? When running with `polkadot-launch` using `mmr-polkadot-v0.9.27` branch everything seems to work fine.

Wizdave97:
> @Wizdave97 we've been trying to fix this for some time with @vmarkushin and unfortunately we are failing to do so. So far we can see that the `Event` type defined on `pallet-ibc` does not match the one generated in docker through `subxt`. We've tried cleaning docker cache but that did not help the problem. Can you take a look at this, to understand what are the discrepancies? When running with `polkadot-launch` using `mmr-polkadot-v0.9.27` branch everything seems to work fine.
I'll take a look into it.

.github/workflows/test.yml
```diff
@@ -51,6 +51,13 @@ jobs:
             pkg-config \
             libssl-dev
 
+          # Get yarn from
+          curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
```
blasrodri, line 55: rm sudo

Vulcan1776:
Task linked: [NX-22 Fix a problem of running the polkadot-launch JS script in centauri repo](https://app.clickup.com/t/20465559/NX-22)

Wizdave97 (approved):

blasrodri (approved):

blasrodri:
LFG

PR #104, opened 2022-11-06
some spelling and assets in ping 

dzmitry-lahoda:
- assuming `Invalid counterparty port` seems wrong error for error in routing, assuming routing routes always correct
- used port str parse instead to_vec in one place as it is in other
- some spelling

### DIFF
--- a/contracts/pallet-ibc/ping/src/lib.rs
+++ b/contracts/pallet-ibc/ping/src/lib.rs
@@ -24,7 +24,7 @@ use ibc::{
 	},
 	signer::Signer,
 };
-use ibc_primitives::{port_id_from_bytes, CallbackWeight, IbcHandler, SendPacketData};
+use ibc_primitives::{CallbackWeight, IbcHandler, SendPacketData};
 use sp_std::{marker::PhantomData, prelude::*};
 // Re-export pallet items so that they can be accessed from the crate namespace.
 pub use pallet::*;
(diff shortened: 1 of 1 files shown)


blasrodri (dismissed):

contracts/pallet-ibc/ping/src/lib.rs
```diff
@@ -182,6 +181,11 @@ impl<T: Config + Send + Sync> Module for IbcModule<T> {
 		version: &Version,
 		counterparty_version: &Version,
 	) -> Result<Version, Ics04Error> {
+		assert_eq!(
+			&PortId::from_str(PORT_ID).expect("PORT_ID is static and valid; qed"),
+			port_id,
+			"routing is correct; qed"
+		);
```
seunlanlege, line 188: we shouldn't panic in the substrate runtime

seunlanlege (changes requested):

contracts/pallet-ibc/ping/src/lib.rs
seunlanlege: also this assertion is incorrect

dzmitry-lahoda: I have replaced assert with error.

dzmitry-lahoda: image
it was this error before. I made different message because `port_id` should be my static port_id? it is not the rule that counter party port will be same as my port id?

dzmitry-lahoda: Counterparty.port_id is is port to which I will send packets, but `port_id` as parameter to method is my port id. I do not see error in assertion text.

dzmitry-lahoda: 🤔

dzmitry-lahoda: after reading more and oding - assertion is correct.

dzmitry-lahoda: it was incorrect, and i fixed it.

Wizdave97 (dismissed):

Wizdave97:
@dzmitry-lahoda  fix conflicts.

dzmitry-lahoda:
too many conflicts, easier to do on top again

PR #110, opened 2022-11-09
move tendermint deps to workspace level when possible

blasrodri:
As a request from Informal Systems, tendermint related deps are being moved to the root Cargo toml root.

Alternative solution is to create a meta-crate. 
```
Move tendermint and ibc dependencies to a higher level .toml file. So we can bump these versions more easily.
Currently, tendermint dependencies are spread across multiple places, difficult to track.
Soares: [Workspaces - The Cargo Book](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacedependencies-table) 
Blas: can do.
MZ: Do you want a metacrate, uncertain how it will work.
David: Yes, a metacrate should work. We need tendermint-proto and tendermint-rs.
MZ: Ok, for follow-up discussion, would like to understand all your tm-rs dependencies.
```

### DIFF
--- a/contracts/pallet-ibc/Cargo.toml
+++ b/contracts/pallet-ibc/Cargo.toml
@@ -30,8 +30,8 @@ sp-trie = { git = "https://github.com/paritytech/substrate", branch = "polkadot-
 cumulus-primitives-core  = { default-features = false, git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.27" }
 # ibc
 ibc-proto = { path = "../../ibc/proto", default-features = false }
-tendermint = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8", optional = true, default-features = false }
-tendermint-proto = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8", default-features = false }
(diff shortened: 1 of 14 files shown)


contracts/pallet-ibc/Cargo.toml
```diff
@@ -30,8 +30,8 @@ sp-trie = { git = "https://github.com/paritytech/substrate", branch = "polkadot-
 cumulus-primitives-core  = { default-features = false, git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.27" }
 # ibc
 ibc-proto = { path = "../../ibc/proto", default-features = false }
-tendermint = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8", optional = true, default-features = false }
-tendermint-proto = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8", default-features = false }
+tendermint = { git = "https://github.com/composableFi/tendermint-rs", rev = "5a74e0f8da4d3dab83cc04b5f1363b018cf3d9e8", optional = true, default-features = false } # cannot be defined as optional in workspace
```
blasrodri, line 33: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacedependencies-table

seunlanlege (approved):

PR #113, opened 2022-11-10
Make core relayer interfaces more generic

### LINKED ISSUE #112: Remove `sp*` types from public interfaces
As part of making hyperspace purely chain agnostic, it would be great to remove substrate dependencies from public interfaces.

From our discussion with the Informal team
```
FS: `sp-std` and `sp-core`
These types are Substrate-specific, they should not appear in chain-agnostic interfaces.
Blas: Agree, we can abstract those away.
David: We can limit that exposure to Parachain-specific implementations.
```

Wizdave97:
closes #112


hyperspace/core/src/chain.rs
```diff
@@ -135,6 +140,7 @@ impl From<String> for AnyError {
 #[async_trait]
 impl IbcProvider for AnyChain {
 	type FinalityEvent = AnyFinalityEvent;
+	type TransactionId = AnyTransactionId;
```
blasrodri, line 143: we'll need to update the docs :P

blasrodri (dismissed):

hyperspace/core/src/chain.rs
Wizdave97: Yeah

seunlanlege (approved):

Issue #122, opened 2022-11-14
E2E Monitoring Event Updates


scglenn:
We need port/channel information in the emitted events in order to monitor events properly, TransferInitiated, TransferReceived, TransferCompleted, TransferFailed.

We also need to update/regenerate the javascript library composablejs-types once these events have been updated.

Once that is completed, my team can continue to perform additional e2e testing on the nodes+relayer when performing ibc transfers

PR #132, opened 2022-11-17
Add channel id to ics20 events

Wizdave97:
Closes https://github.com/ComposableFi/centauri/issues/122
Add more metadata to ibc events.


blasrodri (commented):
Can we have a test ensuring that the event is being sent with the right metadata?

Wizdave97:
> Can we have a test ensuring that the event is being sent with the right metadata?

I don't think that's necessary,  the channel data is always available at this point, so there can't be any errors.

blasrodri (approved):

Issue #125, opened 2022-11-16
implement create connection for cosmos chains


oshorefueled:
To add `create connections` for cosmos chains the following need to be satisfied. 

- [ ] implement `ibc_events` method
- [ ] implement `finality_notifications` method
- [ ] implement `query_client_state`
- [ ] implement `query_latest_ibc_events` method
- [ ] estimate_weight: return 0 from estimate weight so that tx `submit` is immediately processed for cosmos. 

Part of this work was already started on #103, implementations from the PR will be reused.

akc2267:
Hey @oshorefueled, now that https://github.com/ComposableFi/centauri/pull/136 is merged, is this ticket complete?

oshorefueled:
yes it is.

PR #114, opened 2022-11-10
Cosmos support: Implement CosmosClientConfig & CosmosClient

Farhad-Shabani:
Closes #94 

## Steps taken
- Initialized packages for Cosmos support
- Ported necessary traits, structs and methods (Most of them not implemented yet for `CosmosChainClient`)
- Implemented `account_id` from the `KeyProvider` trait. 
- Implemented `submit` from the `Chain` trait
- Implemented `query_client_id_from_tx_hash` from the `IbcProvider` trait

**Note**: There are some other methods required for running setup_clients(). These can be implemented in the next PR.

PR #103, opened 2022-11-04
Cosmos<>Cosmos through Hyperspace as a groundwork for Comsos<>Parachain

Farhad-Shabani:
Closes #94

## Description
With regard to issue #93 and as a very first step towards supporting Cosmos and Parachain by each of the Hyperspace and Hermes teams, This PR attempts to scaffold the necessary functionality for relaying some test packets between Cosmos chains through Hyperspace. Basically, it's laying down the foundation to achieve below objectives: 
- Address light client matters
- Have more generic designs to pave the development process for Cosmos<>Parachain 
- Apply lessons learned from Hyperspace and Hermes experience on both ends, including WIP for [Hermes v1.5](https://github.com/informalsystems/hermes/pull/2478)

PR #136, opened 2022-11-23
create connection on cosmos chains

oshorefueled:
This PR is based off PR #114 and resolves issue #125.


hyperspace/cosmos/src/utils.rs
```diff
@@ -0,0 +1,485 @@
+use crate::error::Error;
```
Farhad-Shabani, line 1: I think most of the necessary functions to extract/filter/converts events exist in the following sections, in particular in the [ics07-tendermint/src/events.rs](https://github.com/ComposableFi/centauri/blob/master/light-clients/ics07-tendermint/src/events.rs). But we have to refactor it and import some others from `ibc-module` so that it can be called in different places for the methods of `IbcProvider` trait  

- [ibc module events](https://github.com/ComposableFi/centauri/blob/master/ibc/modules/src/events.rs
)
- [ics02_client/events
](https://github.com/ComposableFi/centauri/blob/master/ibc/modules/src/core/ics02_client/events.rs
)
- [ics03_connection/events](https://github.com/ComposableFi/centauri/blob/master/ibc/modules/src/core/ics03_connection/events.rs
)
- [ics04_channel/events](https://github.com/ComposableFi/centauri/blob/master/ibc/modules/src/core/ics04_channel/events.rs)

hyperspace/cosmos/src/lib.rs
```diff
@@ -68,6 +73,8 @@ pub struct CosmosClient<H> {
 pub struct CosmosClientConfig {
 	/// Chain name
 	pub name: String,
+	/// peer id for fetching light blocks through rpc io
+	pub peer_id: tendermint::node::Id,
```
Farhad-Shabani, line 77: The `peer_id` should not be required as a configuration field, as it can be obtained via the rpc address. Check [here](https://github.com/informalsystems/hermes/blob/027578f04f6b53faacca6fa2bdbb9018288ee7dc/crates/relayer/src/chain/cosmos.rs#L1835) please

Wizdave97:
You should rebase this PR off the `cosmos-support` branch

Wizdave97:
I don't think the rebase is complete, still seeing code that was reviewed in the previous PR

hyperspace/cosmos/src/utils.rs
oshorefueled: yeah, we'll resolve that in another PR.

hyperspace/cosmos/src/chain.rs
```diff
@@ -27,14 +30,35 @@ where
 		todo!()
 	}
 
-	async fn estimate_weight(&self, _messages: Vec<Any>) -> Result<u64, Self::Error> {
-		todo!()
+	async fn estimate_weight(&self, messages: Vec<Any>) -> Result<u64, Self::Error> {
```
blasrodri, line 33: so that you know, it's standard to put the `_` as a prefix, if the argument is not being used.

hyperspace/cosmos/src/chain.rs
```diff
@@ -27,14 +30,35 @@ where
...
 	}
 
 	async fn finality_notifications(
 		&self,
 	) -> Pin<Box<dyn Stream<Item = <Self as IbcProvider>::FinalityEvent> + Send + Sync>> {
-		todo!()
+		let (ws_client, ws_driver) = WebSocketClient::new(self.websocket_url.clone())
+			.await
+			.map_err(|e| Error::from(format!("Web Socket Client Error {:?}", e)))
+			.unwrap();
```
blasrodri, line 43: do we want to unwrap here?

hyperspace/cosmos/src/provider.rs
```diff
@@ -20,27 +22,95 @@ use ibc_proto::{
...
+
+		Ok((Header{
+			signed_header: latest_light_block.signed_header,
+			validator_set: latest_light_block.validators,
+			trusted_height,
+			trusted_validator_set: trusted_light_block.validators
+		}, update_type))
+	}
+
+	async fn query_tendermint_proof(&self, key: Vec<u8>, height: Height) -> Result<(Vec<u8>, Vec<u8>), Error>{
```
blasrodri, line 83: is it possible to create some type alias, in order to reflect with types what the output of the function is? It's hard for me to picture that with `Vec<u8>`

hyperspace/cosmos/src/provider.rs
```diff
@@ -20,27 +22,95 @@ use ibc_proto::{
...
+		let path = format!("store/{}/key", STORE_KEY).as_str()
```
blasrodri, line 84: small optimization (nit)
I think you can use the `concat!` macro here: https://doc.rust-lang.org/stable/std/macro.concat.html and avoid an allocation

hyperspace/cosmos/src/provider.rs
```diff
@@ -20,27 +22,95 @@ use ibc_proto::{
...
+
+		if !query_res.code.is_ok() {
+			// Fail with response log.
+			// todo()! add response code to error
+			return Err(Error::Custom(format!("failed abci query")));
+		}
+
+		if query_res.proof.is_none() {
+			// Fail due to empty proof
+			return Err(Error::Custom(format!("proof response is empty")));
```
blasrodri, line 102: alternatively you can express this as a `match`

hyperspace/cosmos/src/provider.rs
```diff
@@ -58,11 +128,118 @@ where
...
+									&& event_is_type_channel(&ibc_event)
+								{
+									events.push(ibc_event);
+								}
+							}
+						}
+					},
+					_ => {},
+				}
+				stream::iter(events).map(Ok)
```
blasrodri, line 237: what's this piece trying to achieve?

hyperspace/cosmos/src/utils.rs
```diff
@@ -0,0 +1,485 @@
...
+use ibc::core::ics03_connection::events::{
+	self as ConnectionEvents, Attributes as ConnectionAttributes,
+};
+use ibc::core::ics04_channel::{
+	events::{self as ChannelEvents, Attributes as ChannelAttributes},
+	packet::Packet,
+};
+
+fn extract_block_events(block_events: &HashMap<String, Vec<String>>) -> Vec<IbcEvent> {
+	#[inline]
```
blasrodri, line 21: why do we want to define a function inside a function?

hyperspace/cosmos/src/utils.rs
```diff
@@ -0,0 +1,485 @@
...
+		Ok(attrs) => ChannelEvents::OpenTry::try_from(attrs)
+			.map_err(|e| Error::from(format!("Failed to parse channel open try event {:?}", e))),
+		Err(e) => Err(e),
+	}
+}
+
+pub fn channel_open_ack_try_from_abci_event(
+	abci_event: &AbciEvent,
+) -> Result<ChannelEvents::OpenAck, Error> {
+	match channel_extract_attributes_from_tx(abci_event) {
```
blasrodri, line 216: seems like here you don't need a match, and a `map` would work properly

light-clients/ics07-tendermint/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub mod client_state;
 pub mod consensus_state;
 pub mod error;
 pub mod events;
-mod merkle;
+pub mod merkle;
```
blasrodri, line 36: why this change?

oshorefueled: we need a method from the library to convert tendermint proof to merkle proof

hyperspace/cosmos/src/chain.rs
oshorefueled: yeah, if the we can't connect to the websocket to receive finality notifications, we might as well just panic.

hyperspace/cosmos/src/provider.rs
oshorefueled: wouldn't match result to more lines of code since I'd have to use different match statements for both checks? One checks for code and the other for proof.

blasrodri: The end result is the same, it's just a matter of style. But I would have done it like that. 
Let's assume that query_res is a `QueryResult {proof: Option<T>, code: Result<V>}`
Then you could
```rs
match query_result {
      QueryResult {
           code: Ok(_), ...
      } => return Err(Error::Custom(format!("failed abci query"))),
      QueryResult {
           proof: None, ...
      } => return Err(Error::Custom(format!("failed abci query"))),
}

oshorefueled: wouldn't match result to more lines of code since I'd have to use different match statements for both checks? One checks for code and the other for proof.

oshorefueled: noted

hyperspace/cosmos/src/provider.rs
oshorefueled: This code was gotten from this PR https://github.com/ComposableFi/centauri/pull/103/files#diff-b0380e22a93a5d0cd77b29c49b7bfd1b21aadc9cfc193ecb5d6edc0135e39178R5. It will be refactored on the next PR when we add conversion helper methods to the existing event types.

Issue #93, opened 2022-11-03
Adding Cosmos Support to hyperspace


oshorefueled:
To add a new chain to hyperspace, the chain needs to implement the Chain, *IbcProvider, Keyprovider, Send and Sync traits.* Each implemented method would then be registered in `hyperspace/core/src/chain` by calling them from their respective AnyChain method  in the hyperspace core crate. (Maybe this can be refactored?)

[https://github.com/ComposableFi/centauri/blob/master/hyperspace/core/src/chain.rs#L122](https://github.com/ComposableFi/centauri/blob/master/hyperspace/core/src/chain.rs#L122)

**Possible Refactoring of the AnyChain Implementation:**

Currently Chain, IbcProvider and KeyProvider traits are implemented on the AnyChain enum with the goal of supported chains being defined as variants on the enum. Each supported chain then has to call its equivalent methods from the AnyChain methods in a match case. [https://github.com/ComposableFi/centauri/blob/master/hyperspace/core/src/chain.rs#L126-L151](https://github.com/ComposableFi/centauri/blob/master/hyperspace/core/src/chain.rs#L126-L151)
This would lead to more code to maintain in match cases as more chains are support is added.   

My suggestion is to return the Chain trait from the `into_client` method rather than returning the AnyChain enum type. I think @vmarkushin called it dynamic dispatching in rust. Using that removes the need to register every implemented method by every supported chain in that file. 

**IbcProvider Trait**

The required query methods that need to be implemented to implement the IbcProvider trait can be found here [https://github.com/ComposableFi/centauri/blob/master/hyperspace/primitives/src/lib.rs#L83](https://github.com/ComposableFi/centauri/blob/master/hyperspace/primitives/src/lib.rs#L83)

Some of the methods in the IbcProvider might not translate directly to a cosmos chain. Meaning they might either not be needed to process cosmos messages or the requirement of the method might not be possible in cosmos. Some of these uncertain methods are. 

- `channel_whitelist`
- `expected_block_time`
- `query_client_update_time_and_height`
- `connection_prefix`
- `client_id`
- `connection_id`
- `client_type`
- `query_timestamp_at`
- `query_client_id_from_tx_hash`
- `is_update_required`

**KeyProvider Trait** 

[...]

PR #139, opened 2022-11-23
Cosmos support: Add setup clients

Farhad-Shabani:
This PR is based on #114 to address another part of issue #93
It adds the necessary components to create and setup cosmos clients, including below implementations:
- Listen for block production by Cosmos chains
- Query list of created clients
- Initialize Tendermint client state


hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,415 @@
...
+			vec!["upgrade".to_string(), "upgradedIBCState".to_string()],
+		)
+		.map_err(|e| Error::from(format!("Invalid client state {}", e)))?;
+
+		let target_height = TmHeight::try_from(latest_height_timestamp.0.revision_height)
+			.map_err(|e| Error::from(e.to_string()))?;
+		let client = self
+			.construct_tendermint_client_state::<HostFunctionsManager>(&client_state)
+			.await?;
+		let trusted_block =
```
Wizdave97, line 340: Can't we just derive the consensus state directly from this trusted block?

Farhad-Shabani: Right. That extra verification step shouldn't be needed here. I've refactored it to also comply more with what we have in the `hyperspace/parachain` package. But kept that TmLightClient builder as a separate method `construct_tendemint_light_client()`... just in case needed...

hyperspace/cosmos/src/lib.rs
```diff
@@ -153,24 +181,85 @@ where
...
+		let consensus_state = TmConsensusState::from(trusted_block.signed_header.header);
+
+		Ok((
+			AnyClientState::Tendermint(client_state),
+			AnyConsensusState::Tendermint(consensus_state),
+		))
+	}
+
+	/// Construct a tendermint light client to communicate with the full node
+	pub async fn _construct_tendemint_light_client<HostFunctions: CryptoProvider>(
```
Wizdave97, line 217: We can always import the `ics07-tendermint` crate if we need to do any verification, so I doubt this would ever be needed here.

Farhad-Shabani: Alright, it's removed https://github.com/ComposableFi/centauri/pull/139/commits/4cd48ad8a0b8a2d11504f24e6ac6806aca272da8

Wizdave97 (approved):

oshorefueled (approved):

PR #141, opened 2022-11-23
reverting moving deps into root Cargo file

### LINKED ISSUE #140: cannot use centauri in reproducible builds
# root cause

- pure rust https://github.com/rust-lang/cargo/issues/11192
- same for nix https://github.com/ipetkov/crane/issues/113

# solution

do not use workspace deps until bugs fixed

# alternative 

- be part of monorepo
- fix and release rust :) 

# not alternative

stop doing reproducible builds

blasrodri:
Fixes #140


Wizdave97:
What’s the motivation for reverting it?

blasrodri:
> What’s the motivation for reverting it?

Issue #140
Basically that Dzmitry needs to create a reproducible build and this is preventing him from that

Wizdave97:
Yeah just saw the issue

Wizdave97 (approved):

PR #142, opened 2022-11-24
Remove unsafe unwraps

### LINKED ISSUE #137: Remove unsafe `unwrap` use in `ibc-rs`.
Investigate all places where unwrap is used unsafely including the ones below and modify the code.
https://github.com/ComposableFi/centauri/blob/83cdbd978ab31dd5df1b7c2eff119dc43ef62148/ibc/modules/src/core/ics04_channel/handler/verify.rs#L75
https://github.com/ComposableFi/centauri/blob/83cdbd978ab31dd5df1b7c2eff119dc43ef62148/ibc/modules/src/core/ics04_channel/events.rs#L354 https://github.com/ComposableFi/centauri/blob/83cdbd978ab31dd5df1b7c2eff119dc43ef62148/ibc/modules/src/core/ics04_channel/events.rs#L914

Wizdave97:
closes #137

### DIFF
--- a/ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
+++ b/ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
@@ -100,8 +100,11 @@ where
 				.map(|height| Height::new(height.revision_number, height.revision_height));
 			if proof_bytes.is_some() && consensus_height.is_some() {
 				Some(
-					ConsensusProof::new(proof_bytes.unwrap(), consensus_height.unwrap())
-						.map_err(Error::invalid_proof)?,
+					ConsensusProof::new(
+						proof_bytes.expect("Contains a value"),
+						consensus_height.expect("Contains a value"),
+					)
+					.map_err(Error::invalid_proof)?,
 				)
 			} else {
 				None
--- a/ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
+++ b/ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
@@ -98,13 +98,11 @@ where
 			let consensus_height = msg
 				.consensus_height
 				.map(|height| Height::new(height.revision_number, height.revision_height));
-			if proof_bytes.is_some() && consensus_height.is_some() {
+
+			if let Some((proof_bytes, consensus_height)) = proof_bytes.zip(consensus_height) {
 				Some(
-					ConsensusProof::new(
-						proof_bytes.expect("Contains a value"),
-						consensus_height.expect("Contains a value"),
-					)
-					.map_err(Error::invalid_proof)?,
+					ConsensusProof::new(proof_bytes, consensus_height)
+						.map_err(Error::invalid_proof)?,
 				)
 			} else {
 				None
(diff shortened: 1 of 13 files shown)


ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
```diff
@@ -100,8 +100,11 @@ where
 				.map(|height| Height::new(height.revision_number, height.revision_height));
 			if proof_bytes.is_some() && consensus_height.is_some() {
 				Some(
-					ConsensusProof::new(proof_bytes.unwrap(), consensus_height.unwrap())
-						.map_err(Error::invalid_proof)?,
+					ConsensusProof::new(
+						proof_bytes.expect("Contains a value"),
```
blasrodri, line 104: how are we sure that this doesn't panic?

ibc/modules/src/core/ics23_commitment/commitment.rs
```diff
@@ -30,7 +30,7 @@ pub struct CommitmentRoot {
 
 impl fmt::Debug for CommitmentRoot {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-		let hex = Hex::upper_case().encode_to_string(&self.bytes).unwrap();
+		let hex = Hex::upper_case().encode_to_string(&self.bytes).unwrap_or_default();
```
blasrodri, line 33: why is returning the default a good option?

blasrodri: I know it's debug, but just wondering. It's a way to signal that there's nothing on the stderr/stdout?

ibc/modules/src/core/ics23_commitment/merkle.rs
```diff
@@ -55,8 +55,8 @@ impl<H> From<RawMerkleProof> for MerkleProof<H> {
 			.into_iter()
 			.map(|p| {
 				let mut encoded = Vec::new();
-				prost::Message::encode(&p, &mut encoded).unwrap();
-				prost::Message::decode(&*encoded).unwrap()
+				prost::Message::encode(&p, &mut encoded).expect("Expects valid proof conversion");
+				prost::Message::decode(&*encoded).expect("Expects valid proof conversion")
```
blasrodri, line 59: isn't this potentially part of the same problem? A message that's not correct, that will fail on the decoding piece. And this will cause panics, which is what we're trying to avoid.

blasrodri (commented):
I don't understand why sometimes it's OK to panic, and sometimes not. Would be nice to have an explanation of why it's "safe" in some cases.

ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
Wizdave97: It's contained in an if check before

ibc/modules/src/core/ics23_commitment/merkle.rs
Wizdave97: The comment above the code block explains it, we are converting between the same struct, they are just defined in separate crates because of how ibc_proto is generated.

blasrodri (approved):

ibc/modules/src/core/ics03_connection/msgs/conn_open_ack.rs
nazar-pc: You can use `if let Some(proof_bytes)` so `.expect()` is not needed in the first place, it is more idiomatic in Rust that way.

blasrodri (approved):

PR #145, opened 2022-11-24
CreateClients subcommands should return updated config

blasrodri:
In order to run the relayer on the devenet, we need to provide a programmatic way of retrieveing the updated config once the clients are created.


hyperspace/core/src/command.rs
```diff
@@ -111,7 +119,48 @@ impl Cmd {
 			any_chain_b.name(),
 			client_id_a_on_b
 		);
-		Ok(())
+
+		let parachain_client_a = ParachainClient::try_from(any_chain_a).unwrap();
```
Wizdave97, line 123: We should make this more generic, at this point we only need to update the client id in the config, so why not add a set_client_id method on `AnyConfig`>
So here we just do something like `any_config_a.set_client_id(client_id_a_on_b)`, then we can just jump to 
```let new_config = Config {
			chain_a: any_config_a,
			chain_b: any_config_b,
			core: config.core,
		};

File: hyperspace/core/src/command.rs [hunk 2]
```diff
@@ -63,6 +67,9 @@ pub struct Cmd {
 	/// Channel version
 	#[clap(long)]
 	version: Option<String>,
+	/// Directory to store the updated configurations
+	#[clap(long)]
+	pub config_output_dir: String,
```
Wizdave97, line 72: We might not need this, we should just override the initial config that was passed

hyperspace/core/src/command.rs
```diff
@@ -111,7 +119,48 @@ impl Cmd {
...
+		};
+		let new_config = Config {
+			chain_a: AnyConfig::Parachain(new_parachain_config_a),
+			chain_b: AnyConfig::Parachain(new_parachain_config_b),
+			core: config.core,
+		};
+		Ok(new_config)
 	}
 
 	pub async fn create_connection(&self) -> Result<()> {
```
Wizdave97, line 166: We should use the same techniques for connections and channels.
`set_connection_id`, `set_channel_whitelist`

hyperspace/core/src/command.rs
blasrodri: > set_client_id

Is it that way? I thought it was
```
		config.chain_a.set_client_id(client_id_b_on_a);
		config.chain_b.set_client_id(client_id_a_on_b);
```

Wizdave97: It's the opposite

hyperspace/core/src/command.rs
```diff
@@ -111,19 +112,22 @@ impl Cmd {
 			any_chain_b.name(),
 			client_id_a_on_b
 		);
-		Ok(())
+		config.chain_a.set_client_id(client_id_b_on_a);
+		config.chain_b.set_client_id(client_id_a_on_b);
```
Wizdave97, line 116:
```suggestion
		config.chain_b.set_client_id(client_id_b_on_a);
```

Wizdave97 (approved):
Looks good

hyperspace/core/src/command.rs
blasrodri: oh I didn't see this one - fixed in [924c218](https://github.com/ComposableFi/centauri/pull/145/commits/924c218e65def7953f0cfe1c1ae6032e9a45b307)

Wizdave97:
There's some build errors,

hyperspace/core/src/chain.rs
```diff
@@ -646,4 +646,28 @@ impl AnyConfig {
 				AnyChain::Parachain(ParachainClient::new(config).await?),
 		})
 	}
+
+	pub fn set_client_id(&mut self, client_id: ClientId) {
+		match self {
+			Self::Parachain(chain) => {
+				chain.client_id.replace(client_id);
```
vmarkushin, line 653: nit: you can just do `chain.client_id= Some(client_id);`

vmarkushin (commented):
I think it's better to have this behaviour opt-out or use an additional flag for the output config path. Because the initial config may contain additional information like comments or other fields.

blasrodri:
> I think it's better to have this behaviour opt-out or use an additional flag for the output config path. Because the initial config may contain additional information like comments or other fields.

Initially I added an output argument for the new config. This was more flexible (you could potentially pass the same path), but David suggested to remove it.

vmarkushin:
> > I think it's better to have this behaviour opt-out or use an additional flag for the output config path. Because the initial config may contain additional information like comments or other fields.
> 
> Initially I added an output argument for the new config. This was more flexible (you could potentially pass the same path), but @Wizdave97 suggested to remove it.

I think it's ok to make it optional

blasrodri:
> > > I think it's better to have this behaviour opt-out or use an additional flag for the output config path. Because the initial config may contain additional information like comments or other fields.
> > 
> > 
> > Initially I added an output argument for the new config. This was more flexible (you could potentially pass the same path), but @Wizdave97 suggested to remove it.
> 
> I think it's ok to make it optional

Good idea - will do that.

Wizdave97 (dismissed):

Wizdave97 (approved):

PR #161, opened 2022-12-14
Remove child trie iteration 

Wizdave97:
This change stores 256 of the most recent consensus heights for a specific light client in a bounded btree set, this is allows us to make some preliminary checks when accepting light client updates for block heights that are less than the latest consensus height.

### DIFF
--- a/contracts/pallet-ibc/src/client.rs
+++ b/contracts/pallet-ibc/src/client.rs
@@ -121,16 +121,9 @@ where
 		let consensus_heights = ConsensusHeights::<T>::get(client_id.as_bytes().to_vec());
 		let cs_state = consensus_heights
 			.into_iter()
-			.filter(|(revision_number, revision_height)| {
-				let next_height = Height::new(*revision_number, *revision_height);
-				next_height > height
-			})
-			.take(1)
+			.filter(|next_height| next_height > &height)
 			.next()
-			.map(|(revision_number, revision_height)| {
-				let next_height = Height::new(revision_number, revision_height);
-				self.consensus_state(client_id, next_height).ok()
-			})
+			.map(|next_height| self.consensus_state(client_id, next_height).ok())
 			.flatten();
 
 		Ok(cs_state)
@@ -144,17 +137,10 @@ where
 		let consensus_heights = ConsensusHeights::<T>::get(client_id.as_bytes().to_vec());
 		let cs_state = consensus_heights
 			.into_iter()
-			.filter(|(revision_number, revision_height)| {
-				let next_height = Height::new(*revision_number, *revision_height);
-				next_height < height
-			})
+			.filter(|prev_height| prev_height < &height)
 			.rev()
-			.take(1)
 			.next()
-			.map(|(revision_number, revision_height)| {
-				let next_height = Height::new(revision_number, revision_height);
-				self.consensus_state(client_id, next_height).ok()
-			})
+			.map(|prev_height| self.consensus_state(client_id, prev_height).ok())
 			.flatten();
 		Ok(cs_state)
 	}
@@ -382,12 +368,9 @@ where
 			!client_id.as_str().starts_with("11-beefy")
 		{
 			let mut stored_heights = ConsensusHeights::<T>::get(client_id.as_bytes().to_vec());
-			if let Err(val) =
-				stored_heights.try_insert((height.revision_number, height.revision_height))
-			{
+			if let Err(val) = stored_heights.try_insert(height) {
 				let first = stored_heights
 					.iter()
-					.take(1)
 					.next()
 					.expect("Cannot fail as a value always exists")
 					.clone();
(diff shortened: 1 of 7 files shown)


contracts/pallet-ibc/src/client.rs
```diff
@@ -118,55 +118,45 @@ where
...
-		let mut cs_states = ConsensusStates::<T>::iter_key_prefix(client_id)
-			.map(|(height, cs_state)| {
-				let cs = AnyConsensusState::decode_vec(&cs_state).map_err(|e| {
-					ICS02Error::implementation_specific(format!(
-						"[next_consensus_state]: error decoding consensus state from bytes {}",
-						e
-					))
-				})?;
-				Ok((height, cs))
+		let consensus_heights = ConsensusHeights::<T>::get(client_id.as_bytes().to_vec());
```
blasrodri, line 121: do we need to operate in a `Vac`? Can't we stay on the slice level and avoid an allocation?

contracts/pallet-ibc/src/client.rs
```diff
@@ -118,55 +118,45 @@ where
...
-
-		cs_states.sort_by(|a, b| a.0.cmp(&b.0));
-
-		for cs in cs_states {
-			if cs.0 > height {
-				return Ok(Some(cs.1))
-			}
-		}
+			.take(1)
+			.next()
```
blasrodri, line 129: is this `next relevant`? If the iterator has only one element, and you're mapping over it, would not that be enough?

contracts/pallet-ibc/src/ics23/consensus_states.rs
```diff
@@ -49,45 +48,4 @@ impl<T: Config> ConsensusStates<T> {
...
-
-// #[cfg(test)]
-// mod tests {
-// 	use super::ConsensusStates;
-// 	use crate::mock::*;
-// 	use ibc::core::{ics02_client::height::Height, ics24_host::identifier::ClientId};
-// 	use sp_io::TestExternalities;
-
-// 	#[test]
-// 	fn test_child_trie_prefix_iterator() {
```
blasrodri, line 73: can we keep/make a test for this new data structure?

contracts/pallet-ibc/src/client.rs
```diff
@@ -386,7 +376,29 @@ where
...
+			if let Err(val) =
+				stored_heights.try_insert((height.revision_number, height.revision_height))
+			{
+				let first = stored_heights
+					.iter()
+					.take(1)
+					.next()
+					.expect("Cannot fail as a value always exists")
+					.clone();
+				stored_heights.remove(&first);
```
blasrodri, line 394: do you need the clone above? you're using a reference afterwards

contracts/pallet-ibc/src/client.rs
```diff
@@ -386,7 +376,29 @@ where
...
+		ConsensusStates::<T>::insert(client_id.clone(), height, data);
+		// We do not need this hack for neither beefy nor grandpa clients
+		if !client_id.as_str().starts_with("10-grandpa") &&
+			!client_id.as_str().starts_with("11-beefy")
+		{
+			let mut stored_heights = ConsensusHeights::<T>::get(client_id.as_bytes().to_vec());
+			if let Err(val) =
+				stored_heights.try_insert((height.revision_number, height.revision_height))
+			{
+				let first = stored_heights
```
blasrodri, line 388: don't you want to make a method to get the first available one? Seems like we're using this functionality more than once

blasrodri (approved):
LGTM - minor comments only

contracts/pallet-ibc/src/client.rs
```diff
@@ -118,55 +118,45 @@ where
...
-
-		for cs in cs_states {
-			if cs.0 > height {
-				return Ok(Some(cs.1))
-			}
-		}
+			.take(1)
+			.next()
+			.map(|(revision_number, revision_height)| {
+				let next_height = Height::new(revision_number, revision_height);
```
vmarkushin, line 131: Code duplication: the height was constructed above

contracts/pallet-ibc/src/lib.rs
```diff
@@ -340,6 +340,18 @@ pub mod pallet {
...
 
+	#[pallet::storage]
+	#[allow(clippy::disallowed_types)]
+	/// Consensus heights
+	/// Stored as a tuple of (revision_number, revision_height)
+	pub type ConsensusHeights<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
+		Vec<u8>,
+		BoundedBTreeSet<(u64, u64), frame_support::traits::ConstU32<256>>,
```
vmarkushin, line 351: How about using `Height` instead of `(u64, u64)` to use it's ordering implementation?

vmarkushin (approved):

contracts/pallet-ibc/src/client.rs
Wizdave97: Can't mutate the set using a reference from the set.

contracts/pallet-ibc/src/client.rs
Wizdave97: The get method does not accept a reference

blasrodri (approved):

Issue #160, opened 2022-12-11
improve process of client codegen for subxt


dzmitry-lahoda:
# current:
1. wasm is built
2. node is built
3. node runs wasm
4. tcp call is done to get metadata from node
5. cp of code from subxt produced client

bad because process is many step, large built, needs opening ports, hard to put into nix.

# proposed
1. wasm is built
2. get metadata https://github.com/chevdor/subwasm from wasm blob
3. run subxt to get rs code

good because easy to build, no need to open ports, can easy swap wasm to point to any chain (so can build rs for any target pair of chains - rs per connection), runs well in nix, no copy pasted custom code, code works with any chain (will not fail depending of if chain wss or HTTP to get wasm)

Wizdave97:
Working on this.

PR #166, opened 2022-12-20
Make provision for subxt types generation from runtime wasm blob

Wizdave97:
Closes https://github.com/ComposableFi/centauri/issues/160

Allows generation of subxt types from wasm blob
`cargo run --bin codegen -- --path ./utils/subxt/generated/src --parachain-wasm=<path to parachain wasm file> --relaychain-wasm=<path to relaychain wasm file>`


dzmitry-lahoda:
cc @andor0 may use this too. what ever easier

utils/subxt/codegen/bin/main.rs
```diff
@@ -28,22 +30,40 @@ pub struct Cli {
 	pub para_host: String,
 	#[clap(long, default_value = "9188")]
 	pub para_port: String,
+	/// Path to parachain runtime wasm blob
+	#[clap(long)]
+	pub parachain_wasm: Option<PathBuf>,
```
blasrodri, line 35: Why using option?

blasrodri (approved):

utils/subxt/codegen/bin/main.rs
Wizdave97: A user can still choose to fetch it from a live node, fetching the metadata from wasm blob is optional.

PR #167, opened 2022-12-21
Make signer available in all module callbacks.

### LINKED ISSUE #163: Provide the signer in all module callbacks methods
Allow all callback functions to have access to the relayer who submitted the IBC message.

Wizdave97:
Closes #163


blasrodri (approved):
Looks good. Will need to update this on the monorepo

PR #168, opened 2022-12-28
Fix some issues with ics20 events

Wizdave97:
Fixes a bug with the use of `is_receiver_chain_source` in ics20 event construction 
Use correct ss58 prefix in integration tests
Convert receiver account in ics20 account to a hex string instead of ss58 because ideally two connected parachains will have different ss58 prefixes


contracts/pallet-ibc/src/tests.rs
```diff
@@ -277,7 +277,7 @@ fn on_deliver_ics20_recv_packet() {
 		// Create  a new account
 		let pair = sp_core::sr25519::Pair::from_seed(b"12345678901234567890123456789012");
 		let ss58_address_bytes =
-			ibc_primitives::runtime_interface::account_id_to_ss58(pair.public().0).unwrap();
+			ibc_primitives::runtime_interface::account_id_to_ss58(pair.public().0, 49).unwrap();
```
blasrodri, line 280: why 49 and below 42?

blasrodri (approved):

blasrodri:
Can you add a description of the issue? Helps if we want to start creating a changelog

contracts/pallet-ibc/src/lib.rs
```diff
@@ -589,19 +590,20 @@ pub mod pallet {
...
-				.map_err(|_| Error::<T>::Utf8Error)?;
+			let from = runtime_interface::account_id_to_ss58(
+				account_id_32.into(),
+				<T as frame_system::Config>::SS58Prefix::get(),
+			)
+			.and_then(|val| String::from_utf8(val).map_err(|_| SS58CodecError::InvalidAccountId))
+			.map_err(|_| Error::<T>::Utf8Error)?;
 			let to = match params.to {
 				MultiAddress::Id(id) => {
+					// we convert id to hex string instead of ss58 because destination chain could
```
scglenn, line 601: Are you saying that the ss58 string may not be shared by the same user across both chains?
i.e. the '5abc...' account on chain 1 doesnt belong to the same user on chain 2?
or are you saying the 'from' user could be transferring funds to a different 'to' account?

If we convert one of these 'from' variables to hex then i think that should be consistent for 'to' also, across all the relevant ibc events

scglenn (changes requested):

contracts/pallet-ibc/src/lib.rs
Wizdave97: Converted all to hex string now

scglenn: Shouldnt all the ibc events have matching logic like the "TransferInitiated" event for the from/to addresses being hexified? Or do these from/to hex strings simply get passed around and require no additional converting on events like "TransferFailed" "TokenReceived" etc?

scglenn (dismissed):

Wizdave97:
Yeah the others use the string supplied in the packet data, you can't change those because the packet might come from different chains with different account id formats.

blasrodri (approved):

scglenn:
> Yeah the others use the string supplied in the packet data, you can't change those because the packet might come from different chains with different account id formats.

I'll re run this branch with the monitoring system while performing integration tests, will report back with results 🫡

PR #169, opened 2022-12-29
Client expiry test

### LINKED ISSUE #82: Test client expiry using simnode
We should add a test for client expiry that leverages simnode for forwading time on-chain

Wizdave97:
Closes #82


ibc/modules/src/core/ics02_client/handler/update_client.rs
```diff
@@ -80,12 +80,16 @@ where
 	tracing::debug!("latest consensus state: {:?}", latest_consensus_state);
 
 	let now = ctx.host_timestamp();
-	let duration = now.duration_since(&latest_consensus_state.timestamp()).ok_or_else(|| {
+	let last_update_timestamp =
+		ctx.client_update_time(&client_id, client_state.latest_height()).map_err(|_| {
```
blasrodri, line 84: is there any potential issue on updating the context before checking for expiration?

utils/simnode/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+			frame_system::CheckGenesis::<Self::Runtime>::new(),
+			frame_system::CheckEra::<Self::Runtime>::from(Era::Immortal),
+			frame_system::CheckNonce::<Self::Runtime>::from(nonce),
+			frame_system::CheckWeight::<Self::Runtime>::new(),
+			pallet_transaction_payment::ChargeTransactionPayment::<Self::Runtime>::from(0),
+		)
+	}
+}
+
+#[cfg(test)]
```
blasrodri, line 89: are we running these new tests on CI? 
I guess this https://github.com/ComposableFi/centauri/blob/master/.github/workflows/test.yml#L146 covers the new package right?

Wizdave97: Yes it should

ibc/modules/src/core/ics02_client/handler/update_client.rs
Wizdave97: We are not updating the context, that functions returns the timestamp when the last client update message was processed.

blasrodri (approved):

Issue #95, opened 2022-09-20
Implement mmr verifier in solidity


seunlanlege:
This implementation should be based on: https://github.com/nervosnetwork/merkle-mountain-range

Note: we only need the proof verification implementation not the full library for generating proofs.

Issue #96, opened 2022-09-20
Implement merkle-tree multi-leaf proof verifier in solidity 


seunlanlege:
This lib should be based on: https://github.com/ComposableFi/rs-merkle

PR #175, opened 2023-01-11
added mmr verifier

kkx:
#95 implement mmr verifiier
in the issue suggests implementing only the verifier, in this implementation, the proof generation is also created, which is used to test the verifier. maybe we can get rid of the rest completing the rest of the items and passing the review.

Pending items
- use blake2b as the hash function instead of sha3
- verify the cost of verification, currently it is high according to [this comparison](https://github.com/miguelmota/solidity-mt-vs-mmr)

### DIFF
--- a/contracts/ethereum/src/core/mmr/MMR.sol
+++ b/contracts/ethereum/src/core/mmr/MMR.sol
@@ -427,7 +427,7 @@ library MMR {
         uint256 left,
         uint256 right
     ) public pure returns (bytes32[] memory sortedArr) {
-        if (arr.length == 0) || left >= right {
+        if (arr.length == 0 || left >= right) {
             return arr;
         }
         if (left >= right) return arr;


blasrodri:
> #95 implement mmr verifiier in the issue suggests implementing only the verifier, in this implementation, the proof generation is also created, which is used to test the verifier. maybe we can get rid of the rest completing the rest of the items and passing the review.
> 
> Pending items
> 
> * use blake2b as the hash function instead of sha3
> * verify the cost of verification, currently it is high according to [this comparison](https://github.com/miguelmota/solidity-mt-vs-mmr)

Can you do a bit of research on what avenues are available do decrease the gas costs?

nadeemb53:
@kkx apart from the change in the hasher, we may have to use SCALE instead of ABI codec for compatibility.

Wizdave97:
Also, we'll need the verifier to be a multi-leaf verifier, because we'll be getting batch proofs in beefy.

blasrodri:
> Also, we'll need the verifier to be a multi-leaf verifier, because we'll be getting batch proofs in beefy.

do you have any reference to this?

nadeemb53:
I think @Wizdave97 is referring to #96 
It should be based on [this](https://github.com/ComposableFi/centauri/issues/96)

Wizdave97:
> > Also, we'll need the verifier to be a multi-leaf verifier, because we'll be getting batch proofs in beefy.
> 
> do you have any reference to this?

https://github.com/nervosnetwork/merkle-mountain-range/blob/master/src/mmr.rs#L348

nadeemb53:
Please fix compiler warnings about shadow declarations and function mutability.

contracts/ethereum/package.json
```diff
@@ -26,5 +26,9 @@
     "ts-node": "^10.7.0",
     "typechain": "^8.0.0",
     "typescript": "^4.6.4"
+  },
+  "dependencies": {
+    "@noble/hashes": "^1.1.5",
+    "parity-scale-codec": "^0.8.1"
```
nadeemb53, line 32: just `scale-codec` pkg is fine here

contracts/ethereum/test/mmr.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+
+describe("MMR contract", function () {
+  let mmr: any;
+  let owner: any;
+  beforeEach(async () => {
+    [owner] = await ethers.getSigners();
+    const MMRWrapper = await ethers.getContractFactory("MMRWrapper");
+    mmr = await MMRWrapper.deploy();
+  });
```
nadeemb53, line 18: Recommend using fixtures rather than `beforeEach` to increase test suit performance

contracts/ethereum/test/mmr.test.ts
```diff
@@ -0,0 +1,317 @@
+import { format } from "path";
```
nadeemb53, line 1: Remove this

contracts/ethereum/src/core/mmr/mocks/MMRWrapper.sol
```diff
@@ -0,0 +1,41 @@
...
+import "../MMR.sol";
+
+contract MMRWrapper is MMR {
+
+  bytes public rootHash;
+  bool public verified;
+
+  constructor() {
+    blake2bInstance = Blake2b.init(hex"", 64);
+    scaleCodec = new ScaleCodec();
```
nadeemb53, line 13: Isn't passing the address of an existing contract better than deploying a new one?

contracts/ethereum/src/core/mmr/MMR.sol
```diff
@@ -405,4 +410,40 @@ library MMR {
...
+    /// @param arr array of hashes
+    /// @param left partition left index
+    /// @param right partition right index
+    /// @return sortedArr sorted array
+    function _quickSort(
+        bytes32[] memory arr,
+        uint256 left,
+        uint256 right
+    ) public pure returns (bytes32[] memory sortedArr) {
+        if (arr.length == 0) {
```
blasrodri, line 430: [nit] merge two conditions into one?
```suggestion
        if (arr.length == 0) || left >= right {
```

blasrodri (approved):

nadeemb53 (approved):
LFG 🚀

PR #178, opened 2023-01-17
Light client sync

Wizdave97:
This PR implements a method that the relayer can use to sync light clients to the latest state, in case there's a down time for a period of time.

The light client cannot proceed if it misses any header update at which the authority set changes.

Say the relayer is down for 8 hours and during that time one of bridged chains has changed their authority set 3 times.
Then the light client cannot proceed until it receives all the updates for which the authority set changed in order.

We have two chains A and B:
A has a light client on B and B has a light client on A.
To sync both light clients to the latest state we have:

```rust 
// This checks if chain A's light client on B is up to date with the latest finalized block on A
if !chain_a.is_synced(chain_b) {
  let updates = chain_a.fetch_mandatory_updates(chain_b);
  chain_b.submit(updates);
}

if !chain_b.is_synced(chain_a) {
  let updates = chain_b.fetch_mandatory_updates(chain_a);
  chain_a.submit(updates);
}
```
Also fixes missed header updates from the relaychain which we've noticed happens sometimes while the relayer is actively running.

### DIFF
--- a/algorithms/grandpa/prover/build.rs
+++ b/algorithms/grandpa/prover/build.rs
@@ -22,7 +22,7 @@ static RELAY_URL: Lazy<String> = Lazy::new(|| {
 
 static PARA_URL: Lazy<String> = Lazy::new(|| {
 	let ip = std::env::var("PARA_HOST").unwrap_or_else(|_| "127.0.0.1".to_string());
-	format!("ws://{}:9988", ip)
+	format!("ws://{}:9188", ip)
 });
 
 #[tokio::main]
--- a/algorithms/grandpa/prover/build.rs
+++ b/algorithms/grandpa/prover/build.rs
@@ -22,7 +22,7 @@ static RELAY_URL: Lazy<String> = Lazy::new(|| {
 
 static PARA_URL: Lazy<String> = Lazy::new(|| {
 	let ip = std::env::var("PARA_HOST").unwrap_or_else(|_| "127.0.0.1".to_string());
-	format!("ws://{}:9188", ip)
+	format!("ws://{}:9988", ip)
 });
 
 #[tokio::main]
--- a/algorithms/grandpa/prover/src/lib.rs
+++ b/algorithms/grandpa/prover/src/lib.rs
@@ -196,15 +196,15 @@ where
 	}
 
 	/// Returns a tuple of the finality proof for the given parachain `header_numbers` finalized by
-	/// `latest_finalized_height` and also latest parachain block finalized by the former.
+	/// `latest_finalized_height`.
 	/// If `header_numbers` is empty we only fetch the proof for the latest finalized parachain
 	/// block
 	pub async fn query_finality_proof<H>(
 		&self,
 		previous_finalized_height: u32,
 		latest_finalized_height: u32,
-		mut header_numbers: Vec<T::BlockNumber>,
-	) -> Result<(ParachainHeadersWithFinalityProof<H>, T::BlockNumber), anyhow::Error>
+		header_numbers: Vec<T::BlockNumber>,
+	) -> Result<ParachainHeadersWithFinalityProof<H>, anyhow::Error>
 	where
 		H: Header + codec::Decode,
 		u32: From<<H as Header>::Number>,
@@ -276,9 +276,7 @@ where
 		let mut parachain_headers_with_proof =
 			BTreeMap::<H::Hash, ParachainHeaderProofs>::default();
 
-		let (parachain_headers_with_proof, latest_finalized_para_block) = if header_numbers
-			.is_empty()
-		{
+		let parachain_headers_with_proof = if header_numbers.is_empty() {
 			let parachain_header_bytes = {
 				let key = polkadot::api::storage().paras().heads(&Id(self.para_id));
 				self.relay_client
@@ -307,7 +305,7 @@ where
 					.map_err(|err| anyhow!("Error fetching timestamp with proof: {err:?}"))?;
 			let proofs = ParachainHeaderProofs { state_proof, extrinsic, extrinsic_proof };
 			parachain_headers_with_proof.insert(latest_finalized_hash.into(), proofs);
-			(parachain_headers_with_proof, *para_header.number())
+			parachain_headers_with_proof
 		} else {
 			let change_set = self
 				.relay_client
@@ -361,22 +359,13 @@ where
 				let proofs = ParachainHeaderProofs { state_proof, extrinsic, extrinsic_proof };
 				parachain_headers_with_proof.insert(header.hash().into(), proofs);
 			}
-
-			header_numbers.sort();
-
-			(
-				parachain_headers_with_proof,
-				header_numbers.into_iter().max().expect("Header numbers is not empty"),
-			)
+			parachain_headers_with_proof
 		};
 
-		Ok((
-			ParachainHeadersWithFinalityProof {
-				finality_proof,
-				parachain_headers: parachain_headers_with_proof,
-			},
-			latest_finalized_para_block,
-		))
+		Ok(ParachainHeadersWithFinalityProof {
+			finality_proof,
+			parachain_headers: parachain_headers_with_proof,
+		})
 	}
 
 	/// Returns the finality proof for the given parachain header numbers finalized by the given
@@ -414,14 +403,12 @@ where
 		let previous_finalized_height =
 			validation_data.relay_parent_number.min(previous_finalized_relay_height);
 
-		let (para_headers_with_proof, ..) = self
-			.query_finality_proof(
-				previous_finalized_height,
-				latest_finalized_height,
-				header_numbers,
-			)
-			.await?;
-		Ok(para_headers_with_proof)
+		self.query_finality_proof(
+			previous_finalized_height,
+			latest_finalized_height,
+			header_numbers,
+		)
+		.await
 	}
 
 	/// Queries the block at which the epoch for the given block belongs to ends.
--- a/algorithms/grandpa/prover/src/lib.rs
+++ b/algorithms/grandpa/prover/src/lib.rs
@@ -346,7 +346,10 @@ where
 	}
 
 	/// Queries the block at which the epoch for the given block belongs to ends.
-	pub async fn session_end_for_block(&self, block: u32) -> Result<u32, anyhow::Error> {
+	pub async fn session_start_and_end_for_block(
+		&self,
+		block: u32,
+	) -> Result<(u32, u32), anyhow::Error> {
 		let epoch_addr = polkadot::api::storage().babe().epoch_start();
 		let block_hash = self.relay_client.rpc().block_hash(Some(block.into())).await?;
 		let (previous_epoch_start, current_epoch_start) = self
@@ -358,7 +361,10 @@ where
 			.fetch(&epoch_addr)
 			.await?
 			.ok_or_else(|| anyhow!("Failed to fetch epoch information"))?;
-		Ok(current_epoch_start + (current_epoch_start - previous_epoch_start))
+		Ok((
+			current_epoch_start,
+			current_epoch_start + (current_epoch_start - previous_epoch_start),
+		))
 	}
 
 	/// Returns the session length in blocks
--- a/algorithms/grandpa/prover/src/lib.rs
+++ b/algorithms/grandpa/prover/src/lib.rs
@@ -197,11 +197,11 @@ where
 
 	/// Returns the finality proof for the given parachain header numbers finalized by the given
 	/// relay chain height.
-	pub async fn query_finalized_parachain_headers_with_proof<H>(
+	pub async fn query_finality_proof<H, F>(
 		&self,
-		client_state: &ClientState,
-		mut latest_finalized_height: u32,
-		header_numbers: Vec<T::BlockNumber>,
+		previous_finalized_height: u32,
+		latest_finalized_height: u32,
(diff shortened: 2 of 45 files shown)


algorithms/grandpa/prover/build.rs
```diff
@@ -22,7 +22,7 @@ static RELAY_URL: Lazy<String> = Lazy::new(|| {
 
 static PARA_URL: Lazy<String> = Lazy::new(|| {
 	let ip = std::env::var("PARA_HOST").unwrap_or_else(|_| "127.0.0.1".to_string());
-	format!("ws://{}:9188", ip)
+	format!("ws://{}:9988", ip)
```
blasrodri, line 25: is this change due to the polkadot upgrade? It's odd that the port is changing

algorithms/grandpa/prover/src/lib.rs
```diff
@@ -185,34 +185,8 @@ where
 		T::Hash: From<H::Hash>,
 		H::Number: finality_grandpa::BlockNumberOps,
 		T::BlockNumber: One,
+		F: Fn(&<T as Config>::BlockNumber) -> bool,
```
blasrodri, line 188: why hooking it up through generics and not calling it inside as a normal function?

blasrodri:
I know it's still on draft, but I have a request.
Can you update the description with a simple English/pseudocode explanation on how it will track headers, and in which situations it will trigger actions?

algorithms/grandpa/prover/build.rs
Wizdave97: Nah, just due to my local polkadot launch setup, I'll reverse it when I'm done.

hyperspace/parachain/src/light_client_sync.rs
```diff
@@ -275,9 +275,6 @@ where
 		})
 		.collect::<BTreeSet<_>>();
 
-	// We always insert the latest finalized para height
-	headers_with_events.insert(T::BlockNumber::from(latest_finalized_para_height));
```
blasrodri: why removing this?

Wizdave97: I added a check below to ensure the we only add this when the `previous finalized parachain` height is less than the `latest_finalized_para_height`, if not it could lead to some errors when verifying the client update

algorithms/grandpa/prover/src/lib.rs
```diff
@@ -211,56 +212,40 @@ where
...
+				&*unsafe {
+					unsafe_arc_cast::<_, jsonrpsee_ws_client::WsClient>(
+						self.relay_ws_client.clone(),
+					)
+				},
+				latest_finalized_height,
+			)
+			.await?
+			.ok_or_else(|| {
+				anyhow!("No justification found for block: {:?}", latest_finalized_height)
```
blasrodri, line 235: How is it possible to have a block and no justification?

light-clients/ics10-grandpa/src/tests.rs
```diff
@@ -301,7 +300,8 @@ async fn test_continuous_update_of_grandpa_client() {
 					_ => unreachable!("update handler result has incorrect type"),
 				}
 			},
-			Err(e) => panic!("Unexpected error {:?}", e),
+			// Err(e) => panic!("Unexpected error {:?}", e),
+			_ => {},
```
blasrodri, line 304: Why do you choose to ignore now?

algorithms/grandpa/prover/src/lib.rs
Wizdave97: The API returns an option so we need to handle it.

light-clients/ics10-grandpa/src/tests.rs
Wizdave97: That must have been when I was testing, I'll revert it.

hyperspace/primitives/src/lib.rs
```diff
@@ -370,10 +370,25 @@ pub trait MisbehaviourHandler {
 	) -> Result<(), anyhow::Error>;
 }
 
+/// Provides an interface for syncing light clients to the latest state
+#[async_trait::async_trait]
```
blasrodri, line 374: Can we remove this dependency and add the feature instead?
https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html

blasrodri (dismissed):

hyperspace/primitives/src/lib.rs
Wizdave97: Okay, we'll have to enforce nightly usage in the repo then

hyperspace/core/src/command.rs
```diff
@@ -48,6 +48,11 @@ pub enum Subcommand {
 	CreateConnection(Cmd),
 	#[clap(name = "create-channel", about = "Creates a channel on the specified port")]
 	CreateChannel(Cmd),
+	#[clap(
+		name = "sync-clients",
+		about = "Tries to sync light clients on both chains to the latest"
+	)]
+	SyncClients(Cmd),
```
vmarkushin, line 55: Why can't we just use `relay`?

hyperspace/parachain/src/finality_protocol.rs
```diff
@@ -362,15 +367,43 @@ where
...
+		session_end + 1
+	} else if client_state.latest_relay_height == session_start {
+		session_start + 1
+	} else if client_state.latest_relay_height > session_start &&
+		client_state.latest_relay_height < session_end
+	{
+		client_state.latest_relay_height + 1
+	} else {
+		session_end
+	};
```
vmarkushin, line 383:
```suggestion
	let next_relay_height = if client_state.latest_relay_height >= session_start &&
		client_state.latest_relay_height <= session_end
	{
		client_state.latest_relay_height + 1
	} else {
		session_end
	};
```

But probably it can even be simplified to
```rust
let next_relay_height = if client_state.latest_relay_height == session_end {
    session_end + 1
} else {
    session_end
};
```

hyperspace/primitives/src/lib.rs
blasrodri: @vmarkushin any ideas on this? I’m neutral

vmarkushin: @blasrodri are you sure that we should use an unstable feature instead of a stable crate? :)

vmarkushin: I wouldn't use nightly until it's really needed. In this case, I don't see a neeed for this

hyperspace/parachain/src/finality_protocol.rs
Wizdave97: That doesn't work, it omits in between updates and only allows us send session block updates

Wizdave97: If `client_state.latest_relay_height` is not equal to the session end then we'll miss all other updates because we'll be waiting till the session end, in a case where the session is  4 hours we'll only send an update every four hours.

hyperspace/core/src/command.rs
Wizdave97: The relay command listens for new updates, it doesn't necessarily fetch old updates.

hyperspace/primitives/src/lib.rs
Wizdave97: Yeah let's keep it until it's in stable

hyperspace/parachain/src/finality_protocol.rs
Wizdave97: The second suggestion causes us to omit in between updates, it would be noticeable if the session length is large say a couple hours, you didn't notice it in the test because a session is 10 blocks.

hyperspace/core/src/command.rs
vmarkushin: But isn't this PR about making `relay` fetch old updates?

hyperspace/parachain/src/finality_protocol.rs
vmarkushin: So `prove_finality` doesn't return all blocks in a session (up to the given one)?

hyperspace/core/src/command.rs
blasrodri: I agree. Usage-wise, it's necessary to merge both functionalities.

blasrodri: Are you going to merge the functionality @Wizdave97 ?

hyperspace/parachain/src/finality_protocol.rs
Wizdave97: Not about `prove_finality`, the relayer would wait until the session_end block is finalized until it sends any update, and we don't want that.

hyperspace/core/src/command.rs
Wizdave97: Yeah I can merge both

vmarkushin: @blasrodri @Wizdave97 I mean, we don't need `SyncClients` if `Relay` will be submitting headers from the point it stopped. Can we just call `sync_clients` before the `relay` function if the second won't send blocks from the past?

hyperspace/parachain/src/finality_protocol.rs
vmarkushin: I just thought, we call `prover.session_start_and_end_for_block(client_state.latest_relay_height)`. It means that `client_state.latest_relay_height` will always be in between `session_start` and `session_end`, so the condition is always `true`

hyperspace/core/src/command.rs
Wizdave97: Yeah that's the approach I'm taking

hyperspace/parachain/src/finality_protocol.rs
Wizdave97: We could just use `client_state.latest_relay_height + 1`

hyperspace/parachain/src/light_client_sync.rs
```diff
@@ -0,0 +1,311 @@
...
+				let previous_finalized_height = client_state.latest_relay_height;
+				let session_length = prover.session_length().await?;
+				let (.., session_end_block) =
+					prover.session_start_and_end_for_block(previous_finalized_height).await?;
+				let latest_finalized_height = u32::from(finalized_head.number());
+				let session_changes =
+					(latest_finalized_height - session_end_block) / session_length;
+				// If no session changes have occurred between the last update and the latest
+				// finalized height then the light client is still in sync
+				Ok(!(session_changes >= 1))
```
blasrodri, line 92:
```suggestion
				Ok((session_changes == 0))
```

hyperspace/core/src/command.rs
vmarkushin: Let's remove the command then and just use the function inside `Relay` command

hyperspace/primitives/src/lib.rs
```diff
@@ -370,10 +370,25 @@ pub trait MisbehaviourHandler {
...
+pub trait LightClientSync {
```
blasrodri, line 375: We need to add the docs to reflect the inclusion of this trait.

hyperspace/testsuite/src/lib.rs
```diff
@@ -594,3 +594,32 @@ where
...
+		messages_to_a.clone().into_iter().map(|msg| msg.type_url).collect::<Vec<_>>()
+	);
+	log::info!(target: "hyperspace",
+		"Syncing Client on {} {:?}",
+		chain_b.name(),
+		messages_to_b.clone().into_iter().map(|msg| msg.type_url).collect::<Vec<_>>()
+	);
+	chain_a.submit(messages_to_a).await.unwrap();
+	chain_b.submit(messages_to_b).await.unwrap();
+
```
blasrodri, line 622: shouldn't we check that the client is synced?

hyperspace/testsuite/src/lib.rs
```diff
@@ -594,3 +594,32 @@ where
 	send_packet_and_assert_timeout_on_channel_close(chain_a, chain_b, channel_id).await;
 	handle.abort()
 }
+
+pub async fn client_synchronization_test<A, B>(chain_a: &A, chain_b: &B)
```
blasrodri, line 598: is it possible to have a test in which the client isn't synced, and then it becomes synced? Maybe this is exactly what this test does, and I'm not following it 100%

Wizdave97: Yeah that's what this test does, it sleeps for while after client creation, some sessions pass and the clients become out of sync.

hyperspace/primitives/src/lib.rs
Wizdave97: Done

hyperspace/testsuite/src/lib.rs
Wizdave97: Doing that in the refactored code

Wizdave97:
The sync process is handled by the `relay` function before it starts the main relay loop

blasrodri (approved):
Amazing @Wizdave97

Wizdave97:
> Amazing @Wizdave97

@blasrodri  please approve again, there was a build error

blasrodri (approved):

blasrodri:
@Wizdave97 can you please complete this one?

Wizdave97:
> @Wizdave97 can you please complete this one?

yeah, I've pushed the changes

blasrodri (approved):

PR #180, opened 2023-01-17
SCALE codec tests

nadeemb53:
Signed-off-by: nadeemb53 <nadeemb53@gmail.com>


contracts/ethereum/src/core/node-codec/ScaleCodec.sol
```diff
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.17;
 
-import "../../interfaces/ICodec.sol";
-
-contract ScaleCodec is ICodec {
+contract ScaleCodec {
```
blasrodri, line 4: why not referencing the interface any more?

blasrodri:
Can you try adding some more complex structures (perhaps build them using Rust) and bring them as hex here and ensure that the encoding works also with these?

contracts/ethereum/src/core/node-codec/ScaleCodec.sol
nadeemb53: not needed anymore. It's in use by NodeCodec contract though.

nadeemb53:
> Can you try adding some more complex structures (perhaps build them using Rust) and bring them as hex here and ensure that the encoding works also with these?

as we're dealing with u32 at max, the decoded value would not exceed 2^32-1 (4294967295). But I'll try to add more diverse test cases.

PR #184, opened 2023-01-20
Logging state proof in `query_proof`

nadeemb53:


hyperspace/parachain/src/provider.rs
```diff
@@ -232,7 +232,18 @@ where
...
-
+		let header_hash = self
+			.para_client
+			.rpc()
+			.block_hash(Some(at.revision_height.into()))
+			.await
+			.unwrap();
+		let header = self.para_client.rpc().header(header_hash).await.unwrap().unwrap();
+		let state_root = header.state_root();
+		// log the header_hash, header and state_root
```
Wizdave97, line 243: You should log the key and proof
I don't thimk we need to merge into main though, you can just keep in this branch, since it's just for getting test data

nadeemb53: Polkadot-launch didn't work on my mac (due to maybe the polkadot version upgrade).
I tried emulation of linux/amd-64 but it was too slow 😐
Running the tests here to check the generated logs

blasrodri: > You should log the key and proof I don't think we need to merge into main though, you can just keep in this branch, since it's just for getting test data. @blasrodri what do you think?

Agreed. Worst case, can be added as "trace". But if we don't need to merge, then don't.

nadeemb53:
Closing after getting the required data

PR #187, opened 2023-01-23
implement flat fee for pallet-ibc for ics20

blasrodri:


contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -244,6 +248,15 @@ where
 					destination_channel: packet.destination_channel.to_string().as_bytes().to_vec(),
 				});
 				let packet = packet.clone();
+
+				// Fee _algorithm_:
+				// - if there is a Pool in Pablo, then swap
+				// - if no Pool in Pablo, get percentage fee
+
+				// TODO(vim): handle the case in which the direct pool with USD{T, C} does not exist
+				// through routing. Is it maybe done automatically?
```
blasrodri, line 257: @vimukthi-git can you please comment here?

contracts/pallet-ibc/src/fees.rs
```diff
@@ -0,0 +1,22 @@
...
+	fn pool_exists(pool_id: A::PoolId) -> bool {
+		A::pool_exists(pool_id)
+	}
+
+	pub fn charge_fee() {
+		// Fee _algorithm_:
+		// - if there is a Pool in Pablo, then swap
+		// - if no Pool in Pablo, get percentage fee
+
+		// TODO(vim): handle the case in which the direct pool with USD{T, C} does not exist
```
vimukthi-git, line 19: You need to use the dex-router in that case, it also has the same API as Pablo pallet

blasrodri:
@Wizdave97 can you please review the interaction between the `Ics20ServiceCharge` middleware and how I plan to delegate the fee carging logic with `FeeCharger` based on whether there's a pool, or we fallback on your implementation if there's no pool? (There's a lot missing,  but just want to ensure that the design is correct)

utils/parachain-node/runtime/src/lib.rs
```diff
@@ -722,6 +722,8 @@ impl pallet_ibc::Config for Runtime {
 	type Whitelist = AllowAll;
 	type MemoMessage = MemoMessage;
 	type HandleMemo = ();
+	type PoolId;
+	type Pablo; // TODO: vim - can I get help to construct an impl of this? same for mock
```
blasrodri, line 726: @vimukthi-git

Wizdave97 (commented):
Looking good.

utils/parachain-node/runtime/src/lib.rs
vimukthi-git: You can configure a Pablo instance for the runtime same as in the dex-router here https://github.com/ComposableFi/composable/blob/main/code/parachain/frame/dex-router/src/mock.rs#L238

PR #190, opened 2023-01-27
added merkle multi proof and tests

kkx:


contracts/ethereum/src/core/MerkleMultiProof/mocks/MerkleMultiProofWrapper.sol
```diff
@@ -0,0 +1,15 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.17;
+
+import "../MerkleMultiProof.sol";
+import "hardhat/console.sol";
```
nadeemb53, line 5: not needed here

contracts/ethereum/src/core/MerkleMultiProof/MerkleMultiProof.sol
```diff
@@ -0,0 +1,156 @@
+pragma solidity ^0.8.17;
+
+// SPDX-License-Identifier: Apache2
+struct Node {
+    uint256 k_index;
+    bytes32 node;
+}
+
+library MerkleMultiProof {
```
blasrodri, line 9: is this coming from Polytope? if so, credit their work

nadeemb53 (approved):

PR #192, opened 2023-01-29
Add parity-scale-codec feature for ibc and update pallet-ibc

DaviRain-Su:


Wizdave97:
Hello @DaviRain-Su, Thank you for the work, but we would not like to support scale codec for our IBC protocol since it doesn't offer massive improvements over protobuf encoding.

DaviRain-Su:
> Hello @DaviRain-Su, Thank you for the work, but we would not like to support scale codec for our IBC protocol since it doesn't offer massive improvements over protobuf encoding.

If you support codec, you can directly connect Height portid, channelId, connectionId, etc., and use it directly in the substrate instead of encoding it into vec<u8>. I think this is more human-readable, or there are too many operations and type conversions in the middle

Wizdave97:
If we use codec then we have to use codec when implementing our light client in solidity also.

DaviRain-Su:
> If we use codec then we have to use codec when implementing our light client in solidity also, which might require having a more robust scale codec than we currently have in solidity.

It seems that you are currently developing a bridge with Ethereum. According to your reply, protobuf seems to have a solidity implementation, but parity-scale-codec does not.

blasrodri:
> > If we use codec then we have to use codec when implementing our light client in solidity also, which might require having a more robust scale codec than we currently have in solidity.
> 
> It seems that you are currently developing a bridge with Ethereum. According to your reply, protobuf seems to have a solidity implementation, but parity-scale-codec does not.

We are adding the parity scale codec for it

DaviRain-Su:
nice cool

Blas Rodriguez Irizar ***@***.***> 于2023年2月14日周二 21:00写道：

PR #193, opened 2023-01-31
A few IBC fixes

vmarkushin:
- Fix `UpdateClient` event
- Fix `PacketData` and `PrefixedDenom` serialization
- Fix transfer `Acknowledgement` serialization


ibc/modules/src/applications/transfer/acknowledgement.rs
```diff
@@ -13,66 +13,51 @@
...
 
 impl Acknowledgement {
 	pub fn success() -> Self {
-		Self::Success(ACK_SUCCESS_B64.to_vec())
+		Self { result: Some(ACK_SUCCESS_B64.to_string()), error: None }
 	}
 
 	pub fn from_error(err: Error) -> Self {
-		Self::Error(format!("{}: {}", ACK_ERR_STR, err))
+		Self { result: None, error: Some(err.to_string()) }
```
blasrodri, line 40: By changing this aren’t we going to have problems with other implementations? An enum is different to a struct. 
Just worried about introducing a new protocol that nobody else other than pallet ibc would understand.

ibc/modules/src/applications/transfer/acknowledgement.rs
```diff
@@ -13,66 +13,51 @@
...
-	fn as_ref(&self) -> &[u8] {
-		match self {
-			Acknowledgement::Success(b) => b.as_slice(),
-			Acknowledgement::Error(s) => s.as_bytes(),
+	pub fn into_result(self) -> Result<String, String> {
+		match (self.result, self.error) {
+			(Some(r), None) if r == ACK_SUCCESS_B64 => Ok(r),
+			(Some(r), None) => Err(r),
+			(None, Some(e)) => Err(e),
+			_ => unreachable!("invalid response"),
```
blasrodri, line 52: Do we want to panic in the runtime? @Wizdave97

Wizdave97 (approved):

ibc/modules/src/applications/transfer/acknowledgement.rs
Wizdave97: They're  using a struct now in ibc-go

ibc/modules/src/applications/transfer/acknowledgement.rs
Wizdave97, line 52:
```suggestion
			_ => Err("invalid ack".to_string()),
```

Wizdave97 (changes requested):

ibc/modules/src/applications/transfer/acknowledgement.rs
Wizdave97: https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#acknowledgement-envelope
[type Acknowledgement struct {](https://github.com/cosmos/ibc-go/blob/b2fb1192245134607e0e7a294d09d2aa1145f017/modules/core/04-channel/types/channel.pb.go#L402)

blasrodri: True. Go has no enums. I wanted to confirm that we weren’t introducing a conflict

vmarkushin: I can make it an enum, but then serialization will be implemented manually (not through `derive`)

vmarkushin: Actually, no, I can make it an enum

blasrodri (dismissed):

ibc/modules/src/applications/transfer/acknowledgement.rs
```diff
@@ -13,66 +13,49 @@
...
+use serde::{Deserialize, Serialize};
 
 /// A string constant included in error acknowledgements.
 /// NOTE: Changing this const is state machine breaking as acknowledgements are written into state
 pub const ACK_ERR_STR: &str = "error handling packet on destination chain: see events for details";
-pub const ACK_SUCCESS_B64: &[u8] = b"AQ==";
+pub const ACK_SUCCESS_B64: &str = "AQ==";
 
-#[derive(Clone, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
Wizdave97, line 26: @vmarkushin Does this still serialize into a json object?

vmarkushin: Yes, I've checked

vmarkushin: The variant name is the key in JSON:
`Ack::Result` -> `{ "result": "..." }`
`Ack::Error` -> `{ "error": "..." }`

Wizdave97: Okay cool

Wizdave97 (dismissed):

blasrodri (approved):

PR #194, opened 2023-01-31
Update `tendermint` and tendermint client benchmarks

vmarkushin:


blasrodri (approved):

blasrodri:
I’m good with it. Solve the conflict and merge it ☺️

vmarkushin:
Please don't merge until [vmarkushin/ibc-changes](https://github.com/ComposableFi/centauri/tree/vmarkushin/ibc-changes) is merged

contracts/pallet-ibc/src/client.rs
```diff
@@ -95,7 +95,10 @@ where
 		if cfg!(any(test, feature = "runtime-benchmarks")) {
 			"tendermint".to_string()
 		} else {
-			"beefy".to_string()
+			match <T as Config>::LIGHT_CLIENT_PROTOCOL {
+				LightClientProtocol::Beefy => "beefy".to_string(),
```
blasrodri, line 99: why aren't we adding tendermint as a variant in this enum?

contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -19,12 +19,20 @@ use ics11_beefy::{
...
+use sp_core::{crypto::ByteArray, ed25519, H256};
 use sp_runtime::{
 	app_crypto::RuntimePublic,
 	traits::{BlakeTwo256, ConstU32, Header},
 	BoundedBTreeSet, BoundedVec,
 };
+use tendermint::{
+	crypto::{
+		signature::{Error as TendermintCryptoError, Verifier},
+		Sha256,
```
blasrodri, line 31: don't you want to rename this as TendermintSha256? I'm wondering if this could potentially bring a clash if we'd be using another SHA impl?

light-clients/ics07-tendermint/src/lib.rs
```diff
@@ -41,7 +50,9 @@ mod query;
 /// Host functions that allow the light client verify cryptographic proofs in native.
 pub trait HostFunctionsProvider:
 	ics23::HostFunctionsProvider
-	+ tendermint_light_client_verifier::host_functions::CryptoProvider
+	+ Sha256
+	+ MerkleHash
+	+ Verifier
```
blasrodri, line 55: yay

blasrodri (approved):

contracts/pallet-ibc/src/client.rs
Wizdave97: The host client type can either be beefy or grandpa only, the only reason I return the host client type as tendermint in tests and benchmarks, is to skip the block header check that needs to be done when fetching the host consensus state, since parachains cannot fetch their own consensus state we need to verify it from a header but that header needs to be present in frame_system, which is not possible in this context

blasrodri: thanks

blasrodri (approved):

PR #196, opened 2023-02-01
Hyperspace changes to interface 

vmarkushin:
- Change `query_latest_ibc_events` return type
- Load configs for chains from separate files
- Fix client/client/connection creation CLI

### DIFF
--- a/hyperspace/core/src/lib.rs
+++ b/hyperspace/core/src/lib.rs
@@ -50,15 +50,21 @@ where
 {
 	let (mut chain_a_finality, mut chain_b_finality) =
 		(chain_a.finality_notifications().await, chain_b.finality_notifications().await);
+	// `select!` will spawn a new task for each notification, but we need
+	// the notifications to be processed sequentially
+	let lock_a = tokio::sync::Mutex::new(());
+	let lock_b = tokio::sync::Mutex::new(());
 	// loop forever
 	loop {
 		tokio::select! {
 			// new finality event from chain A
 			result = chain_a_finality.next() => {
+				let _guard = lock_a.lock().await;
 				process_finality_event!(chain_a, chain_b, chain_a_metrics, mode, result)
 			}
 			// new finality event from chain B
 			result = chain_b_finality.next() => {
+				let _guard = lock_b.lock().await;
 				process_finality_event!(chain_b, chain_a, chain_b_metrics, mode, result)
 			}
 		}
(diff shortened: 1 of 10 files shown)


hyperspace/core/src/lib.rs
```diff
@@ -50,15 +50,21 @@ where
...
+	// `select!` will spawn a new task for each notification, but we need
+	// the notifications to be processed sequentially
+	let lock_a = tokio::sync::Mutex::new(());
+	let lock_b = tokio::sync::Mutex::new(());
 	// loop forever
 	loop {
 		tokio::select! {
 			// new finality event from chain A
 			result = chain_a_finality.next() => {
+				let _guard = lock_a.lock().await;
```
blasrodri, line 62: why do you need the lock? The loop isn't spawning a new thread.

blasrodri (approved):
LGTM
Can you also update https://github.com/ComposableFi/centauri/blob/master/hyperspace/config.toml with the new updates on how we deal with config? And probably the docs?

hyperspace/core/src/lib.rs
vmarkushin: There is a comment [above](https://github.com/ComposableFi/centauri/blob/f90ab1f02ec35b6727b8a718e93427e170980657/hyperspace/core/src/lib.rs#L53)

Wizdave97: @vmarkushin I went through the docs and code for tokio::select again and it doesn't return until one feature completes, so we don't need the lock

Wizdave97: The handle of the branch that completed first is executed to completion, the rest are dropped https://github.com/tokio-rs/tokio/blob/80ec80165bfa41c197353983edcbeb64bd45a869/tokio/src/macros/select.rs#L558

hyperspace/core/src/lib.rs
```diff
@@ -50,15 +50,21 @@ where
...
 				process_finality_event!(chain_a, chain_b, chain_a_metrics, mode, result)
 			}
 			// new finality event from chain B
 			result = chain_b_finality.next() => {
+				let _guard = lock_b.lock().await;
```
Wizdave97, line 67: This is not needed for the reasons stated above

Wizdave97 (changes requested):

hyperspace/core/src/lib.rs
blasrodri: Yeah he checked yesterday and saw it. Thanks for looking at it too

PR #198, opened 2023-02-06
WIP: split trie proofs

nadeemb53:


blasrodri:
bullish

.github/workflows/test.yml
```diff
@@ -3,7 +3,7 @@ name: Test
 on:
   pull_request:
     branches:
-      - '*'
+      - main
```
blasrodri, line 6: Centauri's main branch is `master`

contracts/pallet-ibc/rpc/src/lib.rs
```diff
@@ -661,8 +661,11 @@ where
 			.client
 			.expect_header(at)
 			.map_err(|_| RpcError::Custom("Unknown header".into()))?;
-		let state_root: Hasher::Out = header.state_root().into();
-		let ibc_proof = generate_trie_proofs(proof, state_root, child_info, keys).unwrap();
+		let state_root = *header.state_root();
+		let ibc_proof = generate_trie_proofs::<<Block::Header as HeaderT>::Hashing>(
+			proof, state_root, child_info, keys,
+		)
+		.unwrap();
```
blasrodri, line 668: please handle the error, we don't panic on the runtime

light-clients/common/src/lib.rs
```diff
@@ -65,25 +73,37 @@ where
...
+	);
+	let verify_path = verify_trie_proof::<LayoutV0<H>, _, _, _>(
+		&root.into(),
+		child_proof,
+		&[(key, Some(&value))],
+	);
+
+	match (verify_root, verify_path) {
+		(Ok(_), Ok(_)) => Ok(()),
+		_ => Err(anyhow!("verification failed")),
```
blasrodri, line 105: is there any value in discriminating which proof failed?

nadeemb53: definitely

contracts/pallet-ibc/rpc/src/lib.rs
```diff
@@ -754,9 +752,7 @@ where
 			proof, state_root, child_info, keys,
 		) {
 			Ok(p) => p,
-			Err(e) => {
-				return Err(e);
-			},
+			Err(e) => return Err(e),
```
blasrodri, line 755: can't you just
```rs
let ibc_proof = match generate_trie_proofs::<<Block::Header as HeaderT>::Hashing>(
proof, state_root, child_info, keys
)?;
```

blasrodri: https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html

light-clients/common/src/lib.rs
```diff
@@ -190,7 +237,7 @@ impl FromStr for RelayChain {
 pub fn decode_timestamp_extrinsic(ext: &Vec<u8>) -> Result<u64, anyhow::Error> {
 	// Timestamp extrinsic should be the first inherent and hence the first extrinsic
 	// https://github.com/paritytech/substrate/blob/d602397a0bbb24b5d627795b797259a44a5e29e9/primitives/trie/src/lib.rs#L99-L101
-	// Decoding from the [2..] because the timestamp inmherent has two extra bytes before the call
+	// Decoding from the [2..] because the timestamp inherently has two extra bytes before the call
```
Wizdave97, line 240:
```suggestion
	// Decoding from the [2..] because the timestamp inherent has two extra bytes before the call
```

Wizdave97:
Let's have a variant of the verify trie proof function
```rust
pub fn compute_proof_root_hash<'a, L, I, K, V>(
	proof: &[Vec<u8>],
	items: I,
) -> Result<<L::Hash as Hasher>::Out, Error<TrieHash<L>, CError<L>>>
	where
		L: TrieLayout,
		I: IntoIterator<Item = &'a (K, Option<V>)>,
		K: 'a + AsRef<[u8]>,
		V: 'a + AsRef<[u8]>,
{
	// Sort items.
	let mut items = items
		.into_iter()
		.map(|(k, v)| (k.as_ref(), v.as_ref().map(|v| v.as_ref())))
		.collect::<Vec<_>>();
	items.sort();

	if items.is_empty() {
		return if proof.is_empty() { Ok(()) } else { Err(Error::ExtraneousNode) }
	}

	// Check for duplicates.
	for i in 1..items.len() {
		if items[i].0 == items[i - 1].0 {
			return Err(Error::DuplicateKey(items[i].0.to_vec()))
		}
	}

	// Iterate simultaneously in order through proof nodes and key-value pairs to verify.
	let mut proof_iter = proof.iter().map(|i| i.as_slice());
	let mut items_iter = items.into_iter().peekable();

	// A stack of child references to fill in omitted branch children for later trie nodes in the
	// proof.
	let mut stack: Vec<StackEntry<L>> = Vec::new();

	let root_node = match proof_iter.next() {
		Some(node) => node,
		None => return Err(Error::IncompleteProof),
	};
	let mut last_entry = StackEntry::<L>::new(root_node, LeftNibbleSlice::new(&[]), false)?;

	let computed_root = loop {
		// Insert omitted value.
		match last_entry.advance_item(&mut items_iter)? {
			Step::Descend(child_prefix) => {
				let next_entry = last_entry.advance_child_index(child_prefix, &mut proof_iter)?;
				stack.push(last_entry);
				last_entry = next_entry;
			},
			Step::UnwindStack => {
				let is_inline = last_entry.is_inline;
				let node_data = last_entry.encode_node()?;

				let child_ref = if is_inline {
					if node_data.len() > L::Hash::LENGTH {
						return Err(Error::InvalidChildReference(node_data))
					}
					let mut hash = <TrieHash<L>>::default();
					hash.as_mut()[..node_data.len()].copy_from_slice(node_data.as_ref());
					ChildReference::Inline(hash, node_data.len())
				} else {
					let hash = L::Hash::hash(&node_data);
					ChildReference::Hash(hash)
				};

				if let Some(entry) = stack.pop() {
					last_entry = entry;
					last_entry.children[last_entry.child_index] = Some(child_ref);
					last_entry.child_index += 1;
				} else {
					if proof_iter.next().is_some() {
						return Err(Error::ExtraneousNode)
					}
					let computed_root = match child_ref {
						ChildReference::Hash(hash) => hash,
						ChildReference::Inline(_, _) =>
							panic!("the bottom item on the stack has is_inline = false; qed"),
					};
					break computed_root
				}
			},
		}
	};

	Ok(computed_root)
}
```

so the membership verification algo becomes
```rust
let child_trie_root = compute_proof_root_hash(&child_trie_proof, &[(key, Some(&value))],)?;

let computed_state_root = compute_proof_root_hash(&child_trie_proof, &[(child_trie_key, Some(&child_trie_root))],)?;

if computed_state_roof !== root {
  return Err()
}

nadeemb53:
Since `StackEntry` struct is private, it's not worth it to implement this variant of verify function to get the computed root.

PR #201, opened 2023-02-09
Update Main Codebase 0.9.34 #CEN-91

obsessed-cake:


frame/merkle-mountain-range/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
```
blasrodri, line 1: https://github.com/paritytech/substrate/blob/master/frame/beefy-mmr/Cargo.toml

frame/merkle-mountain-range/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+name = "pallet-mmr"
```
blasrodri, line 2: what about https://github.com/paritytech/substrate/blob/master/frame/merkle-mountain-range/Cargo.toml

obsessed-cake: hmm... I tried to find this crate here https://github.com/paritytech/substrate/tree/polkadot-v0.9.34/frame
I'll try to use this version.

dzmitry-lahoda:
@andor0 are you working on it now? do yo need help?

obsessed-cake:
We switched to `polkadot-0.9.36` version.
No longer relevant.

PR #202, opened 2023-02-10
ICS20 Memo

Wizdave97:
Update ibc protos to v7.0.0
Implement memo handler for ICS20


ibc/modules/src/applications/transfer/msgs/transfer.rs
```diff
@@ -60,6 +60,8 @@ pub struct MsgTransfer<C = Coin> {
 	/// Timeout timestamp relative to the current block timestamp.
 	/// The timeout is disabled when set to 0.
 	pub timeout_timestamp: Timestamp,
+	/// Memo field
+	pub memo: String,
```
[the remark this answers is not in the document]
Wizdave97: It's specified as a string in the ibc protos

contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -0,0 +1,179 @@
...
+			Version,
+		},
+		ics24_host::identifier::{ChannelId, ConnectionId, PortId},
+		ics26_routing::context::{Module, ModuleCallbackContext, ModuleOutputBuilder},
+	},
+	signer::Signer,
+};
+use sp_runtime::traits::IdentifyAccount;
+
+/// This middleware should be used to wrap ics20 to execute memo
```
blasrodri, line 21: Can you explain why we need to structure this as a middleware?

contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Memo<T, S: Clone + Default + PartialEq + Eq + Debug> {
+	inner: S,
```
blasrodri, line 24: why not constrain `S` to be `Module`?

utils/parachain-node/runtime/src/lib.rs
```diff
@@ -676,6 +676,25 @@ impl DenomToAssetId<Runtime> for IbcDenomToAssetIdConversion {
 	}
 }
 
+#[derive(
+	Debug, codec::Encode, Clone, codec::Decode, PartialEq, Eq, scale_info::TypeInfo, Default,
+)]
+pub struct MemoMessage;
```
vmarkushin, line 682: Why we don't use string in the runtime?

Wizdave97: Encode, Decode and TypeInfo are not implemented for Strings

contracts/pallet-ibc/src/benchmarks/tendermint_benchmark_utils.rs
```diff
@@ -462,6 +462,7 @@ where
 			)
 			.unwrap(),
 			delay_period,
+			host_consensus_state_proof: vec![],
```
blasrodri, line 465: Do these benchmarks don’t really matter? Otherwise we should be able to add a real proof

Wizdave97: We can't add a real proof at the moment, that would require modifying the blockhash storage in frame system somehow, which we can't

Wizdave97: The solution might be to just add the static cost of hashing a block header and an extra storage read to the open_try and open_ack generated weights.
Those are the significant operations that happen when using the host_consensus_state_proof.

blasrodri: Yeah would be nice to have a proxy there

Wizdave97: I I'll write a separate benchmark for the `host_consensus_state` function, I could access frame system, if I use the raw storage API.

blasrodri (approved):

vmarkushin (approved):

PR #208, opened 2023-02-15
docs: add spec for ibc rate limit

blasrodri:


dzmitry-lahoda:
awesome

dzmitry-lahoda:
@blasrodri may you provide simple trait hook/callback for any custom rate limiter to fit into IBC pallet?

so we can set `no operation larger than 1 DOT can happen

dzmitry-lahoda:
thank for writing out design into doc. instead of reading contract code. so i dislike epoch based design. imho there should be efficient continuous design. so that there is no need to write user code based on what time it is.

blasrodri:
> thank for writing out design into doc. instead of reading contract code. so i dislike epoch based design. imho there should be efficient continuous design. so that there is no need to write user code based on what time it is.

So from your perspective is to have a window that's constantly adjusting?

blasrodri:
> @blasrodri may you provide simple trait hook/callback for any custom rate limiter to fit into IBC pallet?
> 
> so we can set `no operation larger than 1 DOT can happen`

this is also a good idea. As far as I understood, we wanted to implement this as a middleware, so that any behavior should be _easily_ encoded

dzmitry-lahoda:
https://github.com/ComposableFi/centauri/pull/225

blasrodri:
@dzmitry-lahoda how would you update the window on O(1)?
- If you store an aggregated number (scalar), then it's harder to update it by removing the older one, as you don't keep track of individual values.

PR #209, opened 2023-02-16
TX payment query API for Dali #CU-8669j3pbp

dzmitry-lahoda:


hyperspace/parachain/src/chain.rs
```diff
@@ -132,13 +137,24 @@ where
...
+		let dispatch_info =
+			TransactionPaymentApiClient::<sp_core::H256, RuntimeDispatchInfo>::query_info(
+				&*self.para_ws_client,
+				extrinsic.into(),
+				None,
+			)
+			.await
+			.map_err(|e| Error::from(format!("Rpc Error From Estimating weight {:?}", e)))?;
+
+		#[cfg(feature = "dali")]
```
blasrodri, line 149: it will be the same for Picasso and composable right?

blasrodri: if so, then we can just add it right now

blasrodri (approved):

hyperspace/parachain/src/chain.rs
dzmitry-lahoda: same

dzmitry-lahoda: just used already existing feature.

dzmitry-lahoda:
As I see tests involving my changes passed (anyway my changes should be no change to these tests, only to dali tests). So merging it to reuse in monorepo.

obsessed-cake (approved):

blasrodri:
@dzmitry-lahoda be patient and wait until CI passes

dzmitry-lahoda:
@blasrodri sure. sorry. will do next time.

PR #218, opened 2023-02-20
Bug fix in packet querying

Wizdave97:
Fixes a subtle bug in packet querying packets.
The previous code was querying the client states used in checking for proof heights wrongly.


blasrodri:
How come our integration tests were working?

Wizdave97:
> How come our integration tests were working?

The client ids are always the same for the parachains in the tests, and they share similar state since they are tracking the same relay chain.

blasrodri (approved):

vmarkushin (approved):

PR #222, opened 2023-02-22
made pallet-ibc operationally ready

dzmitry-lahoda:


faddat:
@dzmitry-lahoda ser

"operational simplicity is not a joke"

<3

blasrodri (approved):

PR #224, opened 2023-02-27
Interface changes and updates before cosmos merge

vmarkushin:
- Add `query_ibc_balance` by asset id
- Add `get_proof_height` (will be used by cosmos)
- Update tendermint
- Fix minor bugs


contracts/pallet-ibc/Cargo.toml
```diff
@@ -30,8 +30,8 @@ sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v
 sp-trie = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36", default-features = false }
 # ibc
 ibc-proto = { path = "../../ibc/proto", default-features = false }
-tendermint = { git = "https://github.com/informalsystems/tendermint-rs", rev = "47e28b50d20138234f7a8b4254da71469f401714", default-features = false } # cannot be defined as optional in workspace
-tendermint-proto = { git = "https://github.com/informalsystems/tendermint-rs", rev = "47e28b50d20138234f7a8b4254da71469f401714", default-features = false }
+tendermint = { git = "https://github.com/informalsystems/tendermint-rs", rev = "e81f7bf23d63ffbcd242381d1ce5e35da3515ff1", default-features = false } # cannot be defined as optional in workspace
```
blasrodri, line 33: is this post dragon berry?

contracts/pallet-ibc/rpc/src/lib.rs
```diff
@@ -132,6 +133,27 @@ pub struct PacketInfo {
 	pub ack: Option<Vec<u8>>,
 }
 
+impl From<RawPacketInfo> for PacketInfo {
+	fn from(info: RawPacketInfo) -> Self {
+		Self {
+			height: info.height.unwrap_or_default(),
```
blasrodri, line 139: does default on None makes sense? Why not implement `try_from` instead?

contracts/pallet-ibc/rpc/src/lib.rs
```diff
@@ -634,12 +653,24 @@ where
 		}
 	}
 
-	fn query_balance_with_address(&self, addr: String) -> Result<Coin> {
+	fn query_balance_with_address(&self, addr: String, asset_id: AssetId) -> Result<Coin> {
 		let api = self.client.runtime_api();
 		let at = BlockId::Hash(self.client.info().best_hash);
-		let denom = format!("{}", self.chain_props.get("tokenSymbol").cloned().unwrap_or_default());
+		let denom = String::from_utf8_lossy(
```
blasrodri, line 659: why not from_utf8 and return error if can't be decoded?

contracts/pallet-ibc/Cargo.toml
vmarkushin: @blasrodri yes, I think this is 0.29. It's the only version that did work with the ibc-go we use

hyperspace/core/src/chain.rs
```diff
@@ -644,6 +654,13 @@ impl Chain for AnyChain {
 			_ => unreachable!(),
 		}
 	}
+
+	async fn get_proof_height(&self, block_height: Height) -> Height {
+		match self {
+			Self::Parachain(chain) => chain.get_proof_height(block_height).await,
+			_ => unreachable!(),
```
blasrodri, line 661: why not Cosmos?

vmarkushin: We don't have cosmos atm

contracts/pallet-ibc/rpc/src/lib.rs
vmarkushin: Yeah I think that's better

hyperspace/core/src/packets/utils.rs
```diff
@@ -205,35 +201,41 @@ pub async fn construct_timeout_message(
 	next_sequence_recv: u64,
 	proof_height: Height,
 ) -> Result<Any, anyhow::Error> {
-	let key = if sink_channel_end.ordering == Order::Ordered {
-		let path = get_key_path(KeyPathType::SeqRecv, &packet);
-		apply_prefix(sink.connection_prefix().into_vec(), path)
+	let path_type = if sink_channel_end.ordering == Order::Ordered {
```
blasrodri, line 204: we don't currently support `Ordered right`?

vmarkushin: I think we should

blasrodri: Ok!

blasrodri (commented):
Can you also add some more docs clarifying the changes on the piece where you swap the counterparty chains? Just so that's easier to follow why and when each chain goes!

hyperspace/core/src/packets/utils.rs
```diff
@@ -151,45 +151,41 @@ pub async fn verify_delay_passed(
...
-				sink.query_client_update_time_and_height(source.client_id(), proof_height).await
+			let actual_proof_height = source.get_proof_height(proof_height).await;
+			if let Ok((sink_client_update_height, sink_client_update_time)) = sink
+				.query_client_update_time_and_height(source.client_id(), actual_proof_height)
+				.await
 			{
 				let block_delay =
 					calculate_block_delay(connection_delay, sink.expected_block_time());
-				if !has_delay_elapsed(
+				has_delay_elapsed(
```
blasrodri, line 181: nice!

blasrodri (approved):

PR #226, opened 2023-02-27
rate limiter in other direction

dzmitry-lahoda:
can optimize or consider other type of structs to be used later after release:)


contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -209,6 +229,18 @@ where
 				// We need to reject transaction amounts that are larger than u128 since we expect
 				// the balance type of the runtime to be a u128; For a U256 to be converted to a
 				// u128 without truncating, the last two words should be zero
+				let msg = Ics20TransferMsg {
```
blasrodri, line 232: do we need to clone? can't we just operate based on references?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -42,6 +42,26 @@ use ibc_primitives::{CallbackWeight, HandlerMessage, IbcHandler};
 use sp_core::crypto::AccountId32;
 use sp_std::marker::PhantomData;
 
+pub type Ics20TransferMsg = ibc::applications::transfer::msgs::transfer::MsgTransfer<
+	ibc::applications::transfer::Coin<ibc::applications::transfer::PrefixedDenom>,
+>;
+
+#[derive(Debug, Clone, Copy)]
+pub enum FlowType {
```
blasrodri, line 50: can you put some comments on what each of these mean?

dzmitry-lahoda: It is 100% same as Extrinsic names which lead to transfers. Assuming that later the date some protocols will transfer too(they could be different settings). So it is not in and out. Imho naming according extinsics origin is good enough docs. Also enum in reference code of cosmos has not comments on it. So there are 2 reasons not to have doc comments.

contracts/pallet-ibc/src/ics20/mod.rs
dzmitry-lahoda: yes and not. until we get generics over lifetime https://blog.rust-lang.org/inside-rust/2022/07/27/keyword-generics.html, could do 4 things, path each filed separately OR move into struct and move out OR make special referene field stuct OR refactor lower layers to work with struct. 

given these options and hard to decide desing imho, better leave as is with smallest amount of code.

dzmitry-lahoda: so in this case best solution is clone. anyway most of cenauri is very clone heavy. so until it all fixed, one small piece is not essential.

contracts/pallet-ibc/src/ics20/mod.rs
dzmitry-lahoda: There could be 3rd way when message is just sent from liquidation engine, it does not call extinsic. so it would not be subject to rate limiter or different setting. so just retaining room for `InternalTransfer` (while on delivery side I do not care so).

obsessed-cake (approved):

blasrodri (approved):

PR #227, opened 2023-02-27
Hyperspace Cosmos

vmarkushin:
Although most of the functionality was implemented, the implementation shouldn't be considered stable.

`LightClientSync` implementation was omitted, because `Chain::finality_notifications` already does it.

CosmosConfig may be extended in the future, but for testing purposes it contains enough parameters.


hyperspace/core/src/chain.rs
```diff
@@ -177,20 +192,28 @@ impl IbcProvider for AnyChain {
...
 			},
-			_ => unreachable!(),
 		}
 	}
 
 	async fn ibc_events(&self) -> Pin<Box<dyn Stream<Item = IbcEvent> + Send + 'static>> {
 		match self {
 			Self::Parachain(chain) => chain.ibc_events().await,
-			_ => unreachable!(),
+			#[cfg(feature = "cosmos")]
```
blasrodri, line 215: Why feature gating?

vmarkushin: I just thought that it's better to have it than not when it will be needed :) The reason is that when hyperspace supports many clients, the end user will want to opt out of some of them to increase compilation times or not waste time installing other chain dependencies.

blasrodri:
@vmarkushin there are some merge conflicts

blasrodri (commented):
Tests on CI are missing with the `cosmos` compilation feature, right?

vmarkushin:
@blasrodri yes, the CI is unable to run the cosmos tests, so they are gated with `#[ignore]`. The `cosmos` feature is enabled by default

blasrodri:
> @blasrodri yes, the CI is unable to run the cosmos tests, so they are gated with `#[ignore]`. The `cosmos` feature is enabled by default

Why is CI unable? Can we fix it?

vmarkushin:
@blasrodri because cosmos tests require a cosmos chain to be running. Do we have it set up?

blasrodri:
> @blasrodri because cosmos tests require a cosmos chain to be running. Do we have it set up?

let's set it up afterwards

hyperspace/core/src/chain.rs
```diff
@@ -630,7 +739,11 @@ impl Chain for AnyChain {
 				use futures::StreamExt;
 				Box::pin(chain.finality_notifications().await.map(|x| x.into()))
 			},
-			_ => unreachable!(),
+			#[cfg(feature = "cosmos")]
+			Self::Cosmos(chain) => {
+				use futures::StreamExt;
+				Box::pin(chain.finality_notifications().await.map(|x| x.into()))
```
obsessed-cake, line 745:
```suggestion
				Box::pin(chain.finality_notifications().await.map(Into::into))
```

hyperspace/cosmos/src/light_client.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+use tendermint_light_client_verifier::{
+	options::Options as TmOptions,
+	types::{Height as TMHeight, LightBlock, PeerId, Status},
+	PredicateVerifier,
+};
+use tendermint_rpc::{Client, HttpClient, Url};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Verified<H> {
```
blasrodri, line 24: what does verified is? can you add some docs on this struct?

hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,1127 @@
...
+		// contained at `to` and will be fetched below by `msg_update_client_header`
+		for height in from.value()..to.value() {
+			let block_results =
+				self.rpc_client.block_results(TmHeight::try_from(height)?).await.map_err(|e| {
+					Error::from(format!(
+						"Failed to query block result for height {:?}: {:?}",
+						height, e
+					))
+				})?;
+			// TODO: handle block_results.begin_block_events, block_results.end_block_events?
```
blasrodri, line 136: what does this TODO mean?

hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,1127 @@
...
+impl<H> IbcProvider for CosmosClient<H>
+where
+	H: Clone + Send + Sync + 'static,
+{
+	type FinalityEvent = FinalityEvent;
+	type TransactionId = TransactionId<Hash>;
+	type AssetId = String;
+	type Error = Error;
+
+	async fn query_latest_ibc_events<C>(
```
blasrodri, line 96: can you break this function into sub functions ?So that it's easier to follow what's goin' on. I find it a bit too long.

hyperspace/cosmos/src/light_client.rs
vmarkushin: It was copied from `hermes` and actually not used anywhere, so I'll just remove it

hyperspace/cosmos/src/provider.rs
vmarkushin: So that's the part I'm not sure about, I'll check how it's done in `hemes`

vmarkushin: Done

hyperspace/cosmos/src/provider.rs
vmarkushin: Done

Farhad-Shabani:
> Looks good to me, would love to have @Farhad-Shabani to check it out!

My pleasure. I'll do my best to have a look into this today. Btw, if this is blocking you, I can check it out later by reviewing the related branch.

blasrodri:
> > Looks good to me, would love to have @Farhad-Shabani to check it out!
> 
> My pleasure. I'll do my best to have a look into this today. Btw, if this is blocking you, I can check it out later by reviewing the related branch.

It’s not blocking us. I’ll do a more thorough review tomorrow and then this will likely be merged. Will certainly appreciate your input too.

hyperspace/cosmos/src/chain.rs
```diff
@@ -0,0 +1,275 @@
...
+			.await
+			.map_err(|e| Error::from(format!("Web Socket Client Error {:?}", e)))
+			.unwrap();
+		tokio::spawn(ws_driver.run());
+		let subscription = ws_client
+			.subscribe(Query::from(EventType::NewBlock))
+			.await
+			.map_err(|e| Error::from(format!("failed to subscribe to new blocks {:?}", e)))
+			.unwrap()
+			.chunks(6);
```
Farhad-Shabani, line 117: Here, It seems that `chunks(6)` were added to process every multiple block together since the height containing proofs differs one from the height with the IBC event to allow you to append `UpdateClientMessage`. Right?

hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,1149 @@
...
+		// query (exclusively) up to `to`, because the proof for the event at `to - 1` will be
+		// contained at `to` and will be fetched below by `msg_update_client_header`
+		let update_headers =
+			self.msg_update_client_header(from, to, client_state.latest_height).await?;
+		let mut block_events = Vec::new();
+		for height in from.value()..to.value() {
+			let ibc_events = self.parse_ibc_events_at(latest_revision, height).await?;
+			block_events.push(ibc_events);
+		}
+		// we don't submit events for the last block, because we don't have a proof for it
```
Farhad-Shabani, line 134: If so, How should we handle the IBC events within the last block of each notification chunk? The system being event-driven, we wouldn't receive another finality notification for that last block, would we? Has it been implemented somewhere that I missed?

hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,1149 @@
...
+				})?;
+				Any { value, type_url: msg.type_url() }
+			};
+			updates.push((update_client_header, events, update_type));
+		}
+		Ok(updates)
+	}
+
+	// TODO: Changed result: `Item =` from `IbcEvent` to `IbcEventWithHeight` to include the
+	// necessary height field, as `height` is removed from `Attribute` from ibc-rs v0.22.0
```
Farhad-Shabani, line 160: This comment does not match anymore.

hyperspace/cosmos/src/provider.rs
```diff
@@ -0,0 +1,1149 @@
...
+		consensus_height: Height,
+	) -> Result<QueryConsensusStateResponse, Self::Error> {
+		let path_bytes = Path::ClientConsensusState(ClientConsensusStatePath {
+			client_id: client_id.clone(),
+			epoch: consensus_height.revision_number,
+			height: consensus_height.revision_height,
+		})
+		.to_string()
+		.into_bytes();
+		let (query_result, proof) = self.query_path(path_bytes.clone(), at, true).await?;
```
Farhad-Shabani, line 261: IIRC, before, while I was implementing this section, `query_path` calls were not returning the proofs, which made me use `query_proof` method instead. Maybe you test it. However, having Tendermint-rs deps updated, the `query_path` method should be fine now. :ok_hand:

Farhad-Shabani (commented):
Awesome job!

Just a few questions came to my mind, but overall everything looks great (specifically for this step). 
I suggest you do the Cosmos<>Cosmos test with the current setup if it's not ready for Parachain<>Cosmos. It just needs to add a `cosmos_cosmos.rs` test parallel to the others with the configs set), run two [Gaia](https://hermes.informal.systems/tutorials/local-chains/index.html) chains using `gm`, and try out the relayer. Certainly, that will give us a better insight into how the relayer behaves in processing each of the IBC event types.

Keep up the good work!

hyperspace/cosmos/src/lib.rs
```diff
@@ -11,3 +11,23 @@
...
+pub mod light_client;
+pub mod provider;
+#[cfg(any(test, feature = "testing"))]
+pub mod test_provider;
+pub mod tx;
+
+pub type TimeoutHeight = Option<Height>;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct HostFunctions;
```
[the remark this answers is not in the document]
vmarkushin: Perhaps, it was added accidentally. I'll remove it, thanks.

hyperspace/cosmos/src/provider.rs
vmarkushin: As you may have noticed, only the `to` field of the finality event is being used in the function, and `from` is used from the counterparty's client height. So we'll always submit all the blocks

hyperspace/cosmos/src/provider.rs
vmarkushin: We still return `IbcEvent` here, so I think we eventually will replace it with `IbcEventWithHeight`

Farhad-Shabani: I thought we could move away with just the `IbcEvent`. 
If so, makes sense.

blasrodri (approved):

blasrodri (approved):

PR #228, opened 2023-03-02
ics20: introduce rate limit

blasrodri:


dzmitry-lahoda:
closing as not planned next months afaik

PR #231, opened 2023-03-11
Move Status from execute to query, fix getting client state when having

misko9:
Deps or DepsMut. Tested against ibc-go 08-wasm unit testing.

### DIFF
--- a/light-clients/ics10-grandpa-cw/src/contract.rs
+++ b/light-clients/ics10-grandpa-cw/src/contract.rs
@@ -5,11 +5,15 @@ use crate::{
 	log,
 	msg::{
 		CheckForMisbehaviourMsg, ClientStateCallResponse, ContractResult, ExecuteMsg,
-		InitializeState, InstantiateMsg, QueryMsg, StatusMsg, UpdateStateMsg,
+		InitializeState, InstantiateMsg, QueryMsg, QueryResponse, StatusMsg, UpdateStateMsg,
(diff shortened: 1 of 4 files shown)


light-clients/ics10-grandpa-cw/src/contract.rs
```diff
@@ -305,10 +299,29 @@ fn process_message(
...
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn query(_deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+	let client_id = ClientId::from_str("08-wasm-0").expect("client id is valid");
 	match msg {
 		QueryMsg::ClientTypeMsg(_) => unimplemented!("ClientTypeMsg"),
 		QueryMsg::GetLatestHeightsMsg(_) => unimplemented!("GetLatestHeightsMsg"),
+		QueryMsg::Status(StatusMsg {}) => {
+			let client_state = get_client_state::<HostFunctions>(deps).unwrap();
```
blasrodri, line 308: Can we handle the error?

misko9: good call, added along with a unit test for it in the wasm light client

blasrodri (approved):

PR #235, opened 2023-03-16
UpdateStateOnMisbehaviour fix: return client state to be saved,

misko9:
otherwise the client will not be frozen.
Alternatively, we could save the client state from within the contract for both UpdateState and UpdateStateOnMisbehaviour. I expect this would be more efficient.


blasrodri (approved):

blasrodri:
@vmarkushin can you take a look?

PR #240, opened 2023-03-21
Publish docker image #CEN-80

obsessed-cake:


.github/workflows/hyperspace-docker-image.yml
```diff
@@ -24,12 +26,16 @@ jobs:
         sudo chown -R $USER:$USER $GITHUB_WORKSPACE
         docker system prune --force --all --volumes
 
+    - name: Install dependencies
+      run: |
```
blasrodri, line 30: Do you need this?

blasrodri (approved):

.github/workflows/hyperspace-docker-image.yml
obsessed-cake: yes, we use `make` below

PR #243, opened 2023-03-22
add certs to hyperspace image

blasrodri:


blasrodri:
Merging this to expedite SRE

PR #247, opened 2023-03-22
Remove client state from wasmvm msg and results

misko9:
Light client contracts will not be passed the client state. They will read the client state as needed from the shared KVStore. An updated client state will also not be returned. The light client contract will be responsible for writing the client state on an update.

This commit includes PR #241

### DIFF
--- a/light-clients/ics10-grandpa-cw/src/msg.rs
+++ b/light-clients/ics10-grandpa-cw/src/msg.rs
@@ -35,13 +35,25 @@ impl Base64 {
 }
 
 #[cw_serde]
+pub struct GenesisMetadata {
+	pub key: Vec<u8>,
+	pub value: Vec<u8>,
+}
+
+#[cw_serde]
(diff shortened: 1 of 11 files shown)


light-clients/ics10-grandpa-cw/src/msg.rs
```diff
@@ -346,17 +327,17 @@ pub struct CheckSubstituteAndUpdateStateMsg {
 
 #[cw_serde]
 pub struct VerifyUpgradeAndUpdateStateMsgRaw {
-	pub client_id: String,
-	pub old_client_state: Bytes,
-	pub upgrade_client_state: Bytes,
-	pub upgrade_consensus_state: Bytes,
+	pub upgrade_client_state: WasmClientState<FakeInner, FakeInner, FakeInner>,
```
blasrodri, line 330: lol, three fake inner

PR #248, opened 2023-03-23
Use macros in core, add Kusama and Rococo Picasso chains

vmarkushin:
Introduce such macro for adding new chains:
image


hyperspace/core/src/macros.rs
```diff
@@ -96,3 +96,797 @@ macro_rules! process_finality_event {
...
+
+			async fn query_connection_end(
+				&self,
+				at: Height,
+				connection_id: ConnectionId,
+			) -> Result<QueryConnectionResponse, Self::Error> {
+				match self {
+					$(
+						$(#[$($meta)*])*
+						Self::$name(chain) => chain
```
blasrodri, line 234: It almost feel like you could create another macro for all these very similar methods

blasrodri (dismissed):

blasrodri (approved):

blasrodri (approved):

PR #256, opened 2023-03-29
Feature-gate Composable Beefy

vmarkushin:


blasrodri (dismissed):
Can you ensure that the CI checks also do a type checking of this one?

PR #257, opened 2023-03-29
Implement `check_substitute_and_update_state` in cw-grandpa contract

vmarkushin:
- Add `check_substitute_and_update_state` in ibc-rs
- Add `status` method and `Status` type with more granular variants of client's status
- Make Grandpa client expirable


blasrodri:
Can we add a test? @vmarkushin

blasrodri (approved):

PR #268, opened 2023-04-11
Fix OpenInitChannel, WriteAcknowledgement handling and ignore events from other clients

vmarkushin:
- Fix usage of client, connection, and channel ids of source/sink chains
- Filter out events that are not related to the clients specified in the config

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -33,6 +33,8 @@ jobs:
 
     - name: check workspace
       run: |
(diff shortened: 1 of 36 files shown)


blasrodri (approved):

blasrodri:
@vmarkushin can we have a test for this?

blasrodri (approved):

blasrodri (approved):

PR #278, opened 2023-04-19
Use on-chain storage instead of off-chain

vmarkushin:
We have seen that sometimes the state doesn't have the right commitment proofs for a sequence. Our theory is that this is because we are using off chain workers to store some data, and that in the event of forks, we can have a consensus failure.

This change aims to fix this, by moving all critical data away from off-chain worker's storage, and bringing it fully on-chain.


blasrodri:
@dzmitry-lahoda this is critical to be included on the runtime upgrade

PR #280, opened 2023-04-19
A bunch of small changes

vmarkushin:
- Add RPC call delays
- Filter packets with zero height and timestamp
- Don't crash on an error
- More chunking and limiting
- Filter "restart required" error


hyperspace/core/src/lib.rs
```diff
@@ -49,12 +50,9 @@ where
 	A: Chain,
 	B: Chain,
 {
-	let (mut chain_a_finality, mut chain_b_finality) =
-		(chain_a.finality_notifications().await, chain_b.finality_notifications().await);
-
 	// If light clients on both chains are not synced then send the old updates and events before
 	// listening for new events
-	if !chain_a.is_synced(&chain_b).await? {
+	while !chain_a.is_synced(&chain_b).await? {
```
blasrodri, line 55: document the edge case that justifies while

hyperspace/core/src/macros.rs
```diff
@@ -21,6 +21,10 @@ macro_rules! process_finality_event {
 			Some(finality_event) => {
 				log::info!("=======================================================");
 				log::info!("Received finality notification from {}", $source.name());
+				let sink_initial_rpc_call_delay = $sink.rpc_call_delay();
+				let source_initial_rpc_call_delay = $source.rpc_call_delay();
+				let mut had_error = false;
```
blasrodri, line 26: add a todo, to improve this later at least

blasrodri (approved):

blasrodri (approved):

PR #289, opened 2023-04-22
made parity version aligned

dzmitry-lahoda:


blasrodri:
why no changes on the toml?

blasrodri (approved):

PR #296, opened 2023-04-26
Fix IBC transfers when using router

vmarkushin:
There is an issue that if one of the cascaded calls in the router fails, the whole packet is considered failed. But since we don't rollback the previous changes (e.g. an ics-20 transfer), the packet state becomes invalid and may cause some weird issues, e.g. allowing to submit the same packet multiple times, receiving more tokens each time (double spending).

It's currently not clear whether it's possible to allow other modules to fail in case we have pre-checked all the sub-calls in the beginning, because then there is a possibility to mark a good packet (from the ics-20 module's point of view) as failed, although, it shouldn't fail in reality, because initially, the process of fee collection and memo execution was not defined in the standard.

### DIFF
--- a/contracts/pallet-ibc/src/ics20/memo.rs
+++ b/contracts/pallet-ibc/src/ics20/memo.rs
@@ -198,5 +198,6 @@ impl<T: Config + Send + Sync, S: Module + Clone + Default + PartialEq + Eq + Deb
 		<T as Config>::HandleMemo::execute_memo(&packet_data, receiver).map_err(|e| {
 			Error::implementation_specific(format!("Failed to execute memo {:?}", e))
 		})?;
+		Ok(())
 	}
 }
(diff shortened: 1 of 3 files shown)


contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -190,3 +183,21 @@ impl<T: Config + Send + Sync, S: Module + Clone + Default + PartialEq + Eq + Deb
...
+		let packet_data: PacketData =
+			serde_json::from_slice(packet.data.as_slice()).map_err(|e| {
+				Error::implementation_specific(format!("Failed to decode packet data {:?}", e))
+			})?;
+		let receiver = <T as Config>::AccountIdConversion::try_from(packet_data.receiver.clone())
+			.map_err(|_| {
+				Error::implementation_specific(format!("Failed to parse receiver account"))
+			})?
+			.into_account();
+		<T as Config>::HandleMemo::execute_memo(&packet_data, receiver).map_err(|e| {
```
blasrodri, line 198: this function doesn't return an `ack`?

blasrodri (approved):

contracts/pallet-ibc/src/ics20/memo.rs
vmarkushin: No, we only get it from the inner transfer module

PR #305, opened 2023-05-02
fix: tendermint client: don't error when current time equals earliest time. 

misko9:
Delay time period is inclusive.

### DIFF
--- a/light-clients/ics07-tendermint/src/client_state.rs
+++ b/light-clients/ics07-tendermint/src/client_state.rs
@@ -221,6 +221,7 @@ impl<H> ClientState<H> {
 	) -> Result<(), Error> {
 		let earliest_time = processed_time + delay_period_time;
 		/// NOTE: delay time period is inclusive, so if current_time is earliest_time, then we return no error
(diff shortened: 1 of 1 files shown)


light-clients/ics07-tendermint/src/client_state.rs
```diff
@@ -220,7 +220,8 @@ impl<H> ClientState<H> {
 		delay_period_blocks: u64,
 	) -> Result<(), Error> {
 		let earliest_time = processed_time + delay_period_time;
-		if current_time.nanoseconds() <= earliest_time {
+		/// NOTE: delay time period is inclusive, so if current_time is earliest_time, then we return no error
```
blasrodri, line 223:
```suggestion
		/// NOTE: delay time period is inclusive, so if current_time is earliest_time, then we return no error
		/// https://github.com/cosmos/ibc-go/blob/9ebc2f81049869bc40c443ffb72d9f3e47afb4fc/modules/light-clients/07-tendermint/client_state.go#L306
```

blasrodri (approved):

PR #308, opened 2023-05-02
Add flat fee to ICS20 Fee pallet

RustNinja:
- introduced `FlatFeeConverter` trait
- add 3 associated type to configure flat fee calculation. 
- updated `process_fee` function to calculate flat fee
- update mock runtime for tests in pallet IBC
- add unit tests for flat fee charing flow


blasrodri:
Amazing job so far @RustNinja!

I haven't reviewed it yet. But a few checks, to be sure that we aren't missing anything:
- we need to charge in and out on Picasso
- on Composable we will **not charge** on a set of whitelisted channels
- if there's no pool, we fallback to the .4% fee
- what happens if the operation fails? Does it have any IBC acknowledgements, or any error handling we need to be aware of? We had an issue with a previous implementation on that area. Can you try to make some tests where things go wrong, and ensure that the ibc flow (acknowledgement, timeouts, etc) isn't broken in any way? Maybe @vmarkushin can actually help you out here.

contracts/pallet-ibc/src/mock.rs
```diff
@@ -216,10 +218,30 @@ impl Config for Test {
...
+	type AssetId = u128;
+	type Balance = u128;
+	fn get_flat_fee(
+		asset_id: Self::AssetId,
+		fee_asset_id: Self::AssetId,
+		fee_asset_amount: Self::Balance,
+	) -> Option<u128> {
+		if asset_id == 3 {
+			return Some(1000)
+			// None
```
blasrodri, line 233:
```suggestion
			// None
```

contracts/pallet-ibc/src/ics20_fee/mod.rs
```diff
@@ -262,10 +286,28 @@ where
...
+					let fee_asset_id = T::FlatFeeAssetId::get();
+					let fee_asset_amount = T::FlatFeeAmount::get();
+					let flat_fee =
+						T::FlatFeeConverter::get_flat_fee(a, fee_asset_id, fee_asset_amount)
+							.unwrap_or_else(|| {
+								// We have ensured that token amounts larger than the max value for
+								// a u128 are rejected in the ics20 on_recv_packet callback so we
+								// can multiply safely. Percent does Non-Overflowing multiplication
+								// so this is infallible
+								percent * packet_data.token.amount.as_u256().low_u128()
```
blasrodri, line 304: we need to charge 10 USD in the value of that token. Is this what we are doing there?

contracts/pallet-ibc/src/ics20_fee/mod.rs
```diff
@@ -262,10 +286,28 @@ where
 		// Send full amount to receiver using the default ics20 logic
 		// We only take the fee charge if the acknowledgement is not an error
 		if ack.as_ref() == Ics20Ack::success().to_string().as_bytes() {
-			// We have ensured that token amounts larger than the max value for a u128 are rejected
-			// in the ics20 on_recv_packet callback so we can multiply safely.
-			// Percent does Non-Overflowing multiplication so this is infallible
-			let fee = percent * packet_data.token.amount.as_u256().low_u128();
+			let asset_id =
```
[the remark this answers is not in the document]
kkast: @blasrodri maybe we can integrate flat fee first and then add whitelisting next? So we can test all is good on that logic. I can create a separate task for whitelisting

contracts/pallet-ibc/src/ics20_fee/mod.rs
kkast: This is the case when there is no flat fee, so it charges the percentage of total

contracts/pallet-ibc/src/ics20_fee/mod.rs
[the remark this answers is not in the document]
blasrodri: Yes

contracts/pallet-ibc/src/ics20_fee/mod.rs
```diff
@@ -81,6 +92,19 @@ pub mod pallet {
 	}
 }
 
+pub trait FlatFeeConverter {
+	type AssetId;
```
kkast, line 96: Is there an issue adding this trait to composable traits? Although, it sort of makes sense to define FlatFeeConverter trait here because it is going to be used here

RustNinja:
> Amazing job so far @RustNinja!
> 
> I haven't reviewed it yet. But a few checks, to be sure that we aren't missing anything:
> 
> * we need to charge in and out on Picasso
> * on Composable we will **not charge** on a set of whitelisted channels
> * if there's no pool, we fallback to the .4% fee
> * what happens if the operation fails? Does it have any IBC acknowledgements, or any error handling we need to be aware of? We had an issue with a previous implementation on that area. Can you try to make some tests where things go wrong, and ensure that the ibc flow (acknowledgement, timeouts, etc) isn't broken in any way? Maybe @vmarkushin can actually help you out here.

Thanks. 

this changes related only to integration of flat fee and contains nothing related to whitelisted channels.

This changes did not bring any new IBC acknowledgements or error handling that you need to be aware.
only one case: when we not able to get asset_id from denom 
because 
`fn from_denom_to_asset_id(denom: &String) -> Result<T::AssetId, Self::Error>;`
return Result.
but than we fallback to the .4% instead of Flat Fee

kkast (approved):

blasrodri (approved):

PR #309, opened 2023-05-05
Flat fee fix for small transferred amount and  non IBC location assets id

RustNinja:
This PR fix two issue : 
- https://github.com/ComposableFi/composable/issues/3564
- https://github.com/ComposableFi/composable/issues/3565

- move `prefixed_coin` up to use adjusted `denom` value to find `asset_id`
- use `min` function to avoid underflow when subtract fee from transferred amount `fee = fee.min(amount);`
- added unit test to cover test case when transferred amount less then flat fee

### DIFF
--- a/contracts/pallet-ibc/src/ics20_fee/mod.rs
+++ b/contracts/pallet-ibc/src/ics20_fee/mod.rs
@@ -105,6 +105,21 @@ pub trait FlatFeeConverter {
 	) -> Option<u128>;
 }
 
+pub struct NonFlatFeeConverter<T>(PhantomData<T>);
+
+impl<T: Config> FlatFeeConverter for NonFlatFeeConverter<T> {
+	type AssetId = T::AssetId;
+	type Balance = T::Balance;
+
+	fn get_flat_fee(
+		_asset_id: Self::AssetId,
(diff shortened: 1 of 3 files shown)


kkast (approved):

contracts/pallet-ibc/src/ics20_fee/mod.rs
```diff
@@ -105,6 +105,21 @@ pub trait FlatFeeConverter {
 	) -> Option<u128>;
 }
 
+pub struct NonFlatFeeConverter<T>(PhantomData<T>);
```
blasrodri, line 108: What’s the advantage over an enum that contains both cases?

RustNinja: There is implementation of FlatFeeConverter in pablo-pallet
https://github.com/ComposableFi/composable/blob/main/code/parachain/frame/pablo/src/lib.rs#L667-L704

but composable-runtime does not depends on on pablo-pallet(that potentially can have enum implementation with both cases).

This NonFlatFeeConverter is a simple mock that can be used by any runtime that depends only from pallet-ibc and only from pallet-ibc.

blasrodri (approved):

PR #329, opened 2023-05-20
Remove Bob from collator selection to fix block time from 24sec to 12sec

RustNinja:
This PR related to https://github.com/ComposableFi/centauri/issues/331
Integration tests has an issue with block time. **24 sec** for block instead expected **12 sec** per block.

The issue with 24 sec and 12 sec as we expected is that `testnet_geneisis` contains two collators nodes : `Alice and Bob`.
Aura provide Bob ability to produce block in his turn. but there was not Bob collator running at all.
So `Bob collator node `skipped his turn and block produced each 24 sec because only `Alice collator` is running for integration tests.


vmarkushin (approved):
Nice, good catch!

blasrodri (approved):
Marvelous

PR #333, opened 2023-05-23
pallet-ibc: report the asset_id on fee collection

blasrodri:


RustNinja (dismissed):

contracts/pallet-ibc/src/ics20_fee/mod.rs
```diff
@@ -418,7 +418,10 @@ where
 			packet.data = serde_json::to_vec(&packet_data).map_err(|_| {
 				Ics04Error::implementation_specific("Failed to receiver account".to_string())
 			})?;
-			Pallet::<T>::deposit_event(Event::<T>::IbcTransferFeeCollected { amount: fee.into() })
+			Pallet::<T>::deposit_event(Event::<T>::IbcTransferFeeCollected {
+				amount: fee.into(),
+				asset_id,
```
kkast, line 423: assset_id result

kkast: should we even allow to go further if asset_id was an err. i think we should throw and error

kkast (changes requested):

RustNinja (approved):

PR #341, opened 2023-06-02
Fix sending mandatory client updates when don't have packets to send

vmarkushin:
+ fix integer underflow when calculating packet timeout
+ enable overflow checks for release

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -229,4 +229,4 @@ substrate-state-trie-migration-rpc = { git = "https://github.com/paritytech//sub
 substrate-wasm-builder = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
 
 [profile.release]
-overflow-checks = false
+overflow-checks = true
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -226,4 +226,7 @@ substrate-build-script-utils = { git = "https://github.com/paritytech//substrate
 substrate-frame-rpc-system = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
 substrate-prometheus-endpoint = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
 substrate-state-trie-migration-rpc = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
-substrate-wasm-builder = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
\ No newline at end of file
+substrate-wasm-builder = { git = "https://github.com/paritytech//substrate.git", branch = "polkadot-v0.9.39" }
+
+[profile.release]
+overflow-checks = true
```
blasrodri, line 232: do you want to enable this for everything? Or just for hyperspace?

blasrodri (dismissed):
Can you add some comments on the place where you fix the underflow? It may be useful for the future

Cargo.toml
vmarkushin: I think for now let's do for everything, it won't hurt

chris-compfi (approved):

PR #342, opened 2023-06-05
Simplify account_id_to_ss58 by preserving return type

mina86:
By returning `Vec<u8>` rather than `String`, account_id_to_ss58 discards information about the returned value, namely about the fact that it’s valid UTF-8 string.  This forces clients to either use unsafe code or perform UTF-8 validation and dealing with an error case (which in practice is dead code).

Change account_id_to_ss58 to return `String` and thus preserving more information about the value.  Clients who wish to operate on `Vec<u8>` can use `String::into_bytes` to get it.


mina86:
@blasrodri if you have a moment could you take a look?

blasrodri (approved):

blasrodri:
Thanks, @mina86

PR #344, opened 2023-06-06
Fix `AcksPath` construction and cosmos `query_recv_packets` query

vmarkushin:


blasrodri (approved):
We need to improve testing to avoid using symmetric data on both chains.

RustNinja (approved):

PR #346, opened 2023-06-08
oak-audit: issue 3

blasrodri:
Returning error instead of Ok when there's an error on execute

### DIFF
--- a/light-clients/ics10-grandpa-cw/src/contract.rs
+++ b/light-clients/ics10-grandpa-cw/src/contract.rs
@@ -134,12 +134,7 @@ pub fn execute(
 	let client = GrandpaClient::<HostFunctions>::default();
 	let mut ctx = Context::<HostFunctions>::new(deps, env);
 	let client_id = ClientId::from_str("08-wasm-0").expect("client id is valid");
-	let result = process_message(msg, client, &mut ctx, client_id.clone());
-	let data = match result {
-		Ok(res) => res,
-		Err(ContractError::Grandpa(e)) => to_binary(&ContractResult::error(e))?,
-		Err(e) => return Err(e),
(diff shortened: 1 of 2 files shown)


light-clients/ics10-grandpa-cw/src/contract.rs
```diff
@@ -134,12 +134,7 @@ pub fn execute(
 	let client = GrandpaClient::<HostFunctions>::default();
 	let mut ctx = Context::<HostFunctions>::new(deps, env);
 	let client_id = ClientId::from_str("08-wasm-0").expect("client id is valid");
-	let result = process_message(msg, client, &mut ctx, client_id.clone());
-	let data = match result {
-		Ok(res) => res,
-		Err(ContractError::Grandpa(e)) => to_binary(&ContractResult::error(e))?,
-		Err(e) => return Err(e),
-	};
+	let data = process_message(msg, client, &mut ctx, client_id.clone())?;
```
vmarkushin, line 137: We need to test this because the 08-wasm module might not expect a grandpa error

blasrodri: Seems like it does as `call()` returns `(ContractResult, error)`

```go
// call calls the contract with the given payload and writes the result to output.
func call[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *ClientState, payload any) (T, error) {
...
}
```

blasrodri: Ran the tests on this new tendermint-cw on the ibc-go branch and tests pass

vmarkushin (approved):

vmarkushin (approved):

PR #350, opened 2023-06-12
oak-audit: fix issues 2 and 12; and fix misbehaviour check

vmarkushin:
Issue 2: "Impossible to report misbehaviors older than 500 blocks"
Instead of storing the last N headers, we now store all the headers for a fixed period of time (`HEADER_ITEM_LIFETIME`), which allows the fisherman to submit misbehavior even if validators reached the previous limit by submitting many blocks at once. Also, previous authority sets are now stored using the same method.

Also, misbehaviour validation was fixed: before the fix, it was only comparing the base blocks (they should have the same hash) and target block (hashes should be different), but it didn't check that these chains are actually diverging (one of them is a fork of another).


contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -232,21 +218,77 @@ pub enum AnyUpgradeOptions {
...
+			}
+		}
+	}
+
+	impl From<AnyClientState> for Any {
+		fn from(value: AnyClientState) -> Self {
+			match value {
+				AnyClientState::Grandpa(value) => Any {
+					type_url: GRANDPA_CLIENT_STATE_TYPE_URL.to_string(),
+					value: value.encode_vec().unwrap(),
```
blasrodri, line 256: these unwraps are safe?

vmarkushin: Yes, the types are fixed and in the test in doesn't panic, so safe

kkast (changes requested):

light-clients/ics10-grandpa-cw/src/context.rs
```diff
@@ -73,37 +72,38 @@ impl<'a, H> Context<'a, H> {
...
 					self.log(&format!("error saving hash to set: {:?}", e));
 				});
 		}
-		if xs.len() > GRANDPA_BLOCK_HASHES_CACHE_SIZE {
-			let drained = xs.drain(0..xs.len() - GRANDPA_BLOCK_HASHES_CACHE_SIZE);
+		let expired = now_ms.saturating_sub(HEADER_ITEM_LIFETIME.as_millis() as u64);
+		let key = hashes.binary_search_by_key(&expired, |(t, _)| *t).unwrap_or_else(identity);
+		let len = hashes.len();
+		let remove_up_to_index = key.min(len.saturating_sub(HEADER_ITEM_MIN_COUNT));
+		if remove_up_to_index < len {
```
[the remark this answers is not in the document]
vmarkushin: In this case, it means that we need to remove up to `remove_up_to_index`. Let's say 
HEADER_ITEM_MIN_COUNT = 5, and current length is x
if x = 6, then len.saturating_sub(HEADER_ITEM_MIN_COUNT) = 1, so we remove only 1 element, to keep the minimum number of headers
if x = 4, then len.saturating_sub(HEADER_ITEM_MIN_COUNT) = 0, so we don't remove anything, because we haven't reached the minimum number of headers

blasrodri (approved):

kkast (approved):

PR #355, opened 2023-06-19
Find forced authorities change in hyperspace

vmarkushin:


blasrodri:
@vmarkushin can we fix merge issues?

PR #356, opened 2023-06-19
Use separate threads for chains

vmarkushin:


hyperspace/core/src/lib.rs
```diff
@@ -53,19 +55,40 @@ where
...
+				chain_a,
+				chain_b_metrics,
+				mode,
+				result,
+				chain_b_finality
+			)
+		}
+	});
+
+	jh1.await??;
```
blasrodri, line 89: we might want to use `join!()` instead here

vmarkushin: Why?

PR #359, opened 2023-06-26
add packet-forward-middleware logic into pallet-ibc pallet

RustNinja:
This PR focused to bring packet-forward-middleware functionality and pass message forward if:
- memo exists
- known assset id

Test / Tests (pull_request) failed because of github ssl 
`curl: (60) SSL: no alternative certificate subject name matches target host name '[objects.githubusercontent.com](http://objects.githubusercontent.com/)'`

Then compose new memo from the prev and call pallet_ibc transfer extrinsic

### DIFF
--- a/contracts/pallet-ibc/src/mock.rs
+++ b/contracts/pallet-ibc/src/mock.rs
@@ -1,5 +1,7 @@
 use crate::{
-	self as pallet_ibc, ics20::SubstrateMultihopXcmHandlerNone, ics20_fee::FlatFeeConverter,
+	self as pallet_ibc,
+	ics20::{IbcModule, SubstrateMultihopXcmHandlerNone},
+	ics20_fee::FlatFeeConverter,
 	routing::ModuleRouter,
 };
 use cumulus_primitives_core::ParaId;
@@ -231,7 +233,7 @@ impl Config for Test {
 	type IbcAccountId = Self::AccountId;
 	type TransferOrigin = EnsureSigned<Self::IbcAccountId>;
 	type RelayerOrigin = EnsureSigned<Self::AccountId>;
-	type HandleMemo = ();
+	type HandleMemo = IbcModule<Test>;
 	type MemoMessage = alloc::string::String;
 	type IsReceiveEnabled = sp_core::ConstBool<true>;
 	type IsSendEnabled = sp_core::ConstBool<true>;
--- a/contracts/pallet-ibc/src/tests.rs
+++ b/contracts/pallet-ibc/src/tests.rs
@@ -3,8 +3,8 @@ use crate::{
 	light_clients::{AnyClientState, AnyConsensusState},
 	mock::*,
 	routing::Context,
-	Any, Config, ConsensusHeights, DenomToAssetId, MultiAddress, Pallet, PendingRecvPacketSeqs,
-	PendingSendPacketSeqs, Timeout, TransferParams, MODULE_ID,
+	Any, Config, ConsensusHeights, DenomToAssetId, Event, MultiAddress, Pallet,
+	PendingRecvPacketSeqs, PendingSendPacketSeqs, Timeout, TransferParams, MODULE_ID,
 };
 use core::time::Duration;
 use frame_support::{
@@ -426,6 +426,142 @@ fn on_deliver_ics20_recv_packet() {
 		let fee = <Test as crate::ics20_fee::Config>::ServiceChargeIn::get() * amt;
 		assert_eq!(balance, amt - fee);
 		assert_eq!(pallet_balance, fee)
+	})
+}
+
+#[test]
+fn on_deliver_ics20_recv_packet_incorrect_memo() {
+	let mut ext = new_test_ext();
+	ext.execute_with(|| {
+		let incorrect_memo = "Incorrect memo".to_string();
+		// Create  a new account
+		let pair = sp_core::sr25519::Pair::from_seed(b"12345678901234567890123456789012");
+		let reciever = AccountId32::new(pair.public().0);
+		let ss58_address =
+			ibc_primitives::runtime_interface::account_id_to_ss58(pair.public().0, 49);
+		frame_system::Pallet::<Test>::set_block_number(1u32);
+		let asset_id =
+			<<Test as Config>::IbcDenomToAssetIdConversion as DenomToAssetId<Test>>::from_denom_to_asset_id(
+				&"PICA".to_string(),
+			)
+			.unwrap();
+		setup_client_and_consensus_state(PortId::transfer());
+
+		let channel_id = ChannelId::new(0);
+		let balance = 100000 * MILLIS;
+
+		// We are simulating a transfer back to the source chain
+
+		let denom = "transfer/channel-1/PICA";
+		let channel_escrow_address =
+			get_channel_escrow_address(&PortId::transfer(), channel_id).unwrap();
+		let channel_escrow_address =
+			<Test as Config>::AccountIdConversion::try_from(channel_escrow_address)
+				.map_err(|_| ())
+				.unwrap();
+		let channel_escrow_address = channel_escrow_address.into_account();
+
+		// Endow escrow address with tokens
+		let _ = <<Test as Config>::NativeCurrency as Currency<
+			<Test as frame_system::Config>::AccountId,
+		>>::deposit_creating(&channel_escrow_address, balance);
+
+		let prefixed_denom = PrefixedDenom::from_str(denom).unwrap();
+		let amt = 1000 * MILLIS;
+		println!("Transferred Amount {}", amt);
+		let coin = Coin {
+			denom: prefixed_denom,
+			amount: ibc::applications::transfer::Amount::from_str(&format!("{:?}", amt)).unwrap(),
+		};
+		let packet_data = PacketData {
+			token: coin,
+			sender: Signer::from_str("alice").unwrap(),
+			receiver: Signer::from_str(&ss58_address).unwrap(),
+			memo: incorrect_memo.clone(),
+		};
+
+		let data = serde_json::to_vec(&packet_data).unwrap();
+		let packet = Packet {
+			sequence: 1u64.into(),
+			source_port: PortId::transfer(),
+			source_channel: ChannelId::new(1),
+			destination_port: PortId::transfer(),
+			destination_channel: ChannelId::new(0),
+			data,
+			timeout_height: Height::new(2000, 5),
+			timeout_timestamp: ibc::timestamp::Timestamp::from_nanoseconds(
+				1690894363u64.saturating_mul(1000000000),
+			)
+			.unwrap(),
+		};
+
+		let msg = MsgRecvPacket {
+			packet,
+			proofs: Proofs::new(
+				vec![0u8; 32].try_into().unwrap(),
+				None,
+				None,
+				None,
+				Height::new(0, 1),
+			)
+			.unwrap(),
+			signer: Signer::from_str(MODULE_ID).unwrap(),
+		};
+
+		let msg = Any { type_url: msg.type_url(), value: msg.encode_vec().unwrap() };
+
+		let account_data = Assets::balance(asset_id, AccountId32::new(pair.public().0));
+		// Assert account balance before transfer
+		assert_eq!(account_data, 0);
+		Ibc::deliver(RuntimeOrigin::signed(AccountId32::new([0; 32])), vec![msg]).unwrap();
+
+		let balance = <<Test as Config>::NativeCurrency as Currency<
+			<Test as frame_system::Config>::AccountId,
+		>>::free_balance(&reciever);
+
+		let pallet_balance =
+			<<Test as Config>::NativeCurrency as Currency<
+				<Test as frame_system::Config>::AccountId,
+			>>::free_balance(&<Test as crate::Config>::FeeAccount::get().into_account());
+		let fee = <Test as crate::ics20_fee::Config>::ServiceChargeIn::get() * amt;
+		assert_eq!(balance, amt - fee);
+		assert_eq!(pallet_balance, fee);
+
+		// let test_event = TestEvent::PalletIbc($event);
+		assert_eq!(
+			System::events()
+				.iter()
+				.filter(|a| {
+					if let RuntimeEvent::Ibc(ibc_event) = &a.event {
+						if let Event::<Test>::ExecuteMemoStarted { memo, .. } = ibc_event {
+							return memo.as_ref().map_or(false, |m| m == &incorrect_memo)
+						}
+					}
+					false
+				})
+				.count(),
+			1
+		);
+
+		assert_eq!(
+			System::events()
+				.iter()
+				.filter(|a| {
+					if let RuntimeEvent::Ibc(ibc_event) = &a.event {
+						if let Event::<Test>::ExecuteMemoIbcTokenTransferFailedWithReason {
+							memo,
+							reason,
+							from,
+						} = ibc_event
+						{
+							return memo == &incorrect_memo && *reason == 0 && from == &reciever
+						}
+					}
+					false
+				})
+				.count(),
+			1
+		);
 	})
 }
 


contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -195,7 +195,7 @@ impl<T: Config + Send + Sync, S: Module + Clone + Default + PartialEq + Eq + Deb
 				Error::implementation_specific(format!("Failed to parse receiver account"))
 			})?
 			.into_account();
-		<T as Config>::HandleMemo::execute_memo(&packet_data, receiver).map_err(|e| {
+		<T as Config>::HandleMemo::execute_memo(packet, &packet_data, receiver).map_err(|e| {
```
blasrodri, line 198: why don't we just send the entire packet, instead of the packet and the packet data?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+use frame_system::RawOrigin;
+use scale_info::prelude::boxed::Box;
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+struct MemoForward {
+	receiver: String,
+	port: String,
+	channel: String,
+	timeout: String,
+	retries: u64,
+	next: Option<Box<MemoForward>>,
```
blasrodri, line 590: would it make sense to have some validation logic to ensure that this isn't a graph that contains a loop?

contracts/pallet-ibc/src/ics20/memo.rs
RustNinja: True.
pass packet and packet_data. to not deserialise two times.
or move all logic(deserialise and AccountIdConversion) from process_memo into <T as Config>::HandleMemo::execute_memo

contracts/pallet-ibc/Cargo.toml
```diff
@@ -139,8 +138,9 @@ std = [
   "sp-finality-grandpa/std",
   "sp-finality-grandpa/std",
   "finality-grandpa/std",
-#  "serde",
+  # "serde",
```
vmarkushin, line 141: Let's just remove it

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+			Ics20Error::implementation_specific("Failed to parse channel ID".to_string())
+		})?;
+
+		//timestamp that current timestamp + 30 minutes
+		let timestamp = <pallet_timestamp::Pallet<T> as frame_support::traits::UnixTime>::now()
+			.as_secs() + 1800;
+
+		let params = crate::TransferParams::<<T as frame_system::Config>::AccountId> {
+			to: account_id,
+			source_channel: channel_id, /* TODO source? should it be destination channel id from
```
vmarkushin, line 669: Yes, in memo you put source channel of the intermediate chain

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+			                             * memo? */
+			timeout: ibc_primitives::Timeout::Offset {
+				timestamp: Some(timestamp), /* TODO ? what timestamp to use. now is current + 30
```
vmarkushin, line 672: This should work fine

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+				                             * minutes */
+				height: None, //TODO ? what height
```
vmarkushin, line 674: Just leave it as None

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+			},
+		};
+
+		let memo = match memo.next {
+			Some(e) => Some(*e),
+			None => None,
+		};
```
vmarkushin, line 681:
```suggestion
		let memo = memo.next.as_deref();
```

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +561,187 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+			let memo_str = serde_json::to_string(&memo).map_err(|_| {
+				Self::emit_memo_execution_failed_event(
+					receiver.clone(),
+					packet_data.memo.clone(),
+					3,
+				);
+				Ics20Error::implementation_specific("failed to serialize memo".to_string())
+			})?;
+			let memo_result =
+				<T as crate::Config>::MemoMessage::from_str(&memo_str).map_err(|_| {
```
vmarkushin, line 694: I guess, it's safe to use the original memo, because it was deserialised successfully

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: I think it's not necessary at the moment. It won't lead to an infinite loop and also it's unlikely to happen if we have validation on front-end side

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: We can't use original memo. we need to remove parent(current chain) because we execute multihop logic reading memo and should put modified memo for next chains

contracts/pallet-ibc/src/ics20/mod.rs
blasrodri: this could be a potential "vulnerability". Not critical in terms of funds being lost, but unnecessary usage of resources, right?

RustNinja: True.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Yes, but the memo is already extracted [here](https://github.com/ComposableFi/centauri/blob/68e96e4aa35f997ac4243d25f067b7aac1208fd0/contracts/pallet-ibc/src/ics20/mod.rs#L786), so I suppose we can use it as is, without [converting](https://github.com/ComposableFi/centauri/blob/68e96e4aa35f997ac4243d25f067b7aac1208fd0/contracts/pallet-ibc/src/ics20/mod.rs#L790-L808) to JSON string and back.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: @blasrodri we always extract `.next` from the memo, so the loop is not possible. Please confirm @RustNinja

RustNinja: we extract next memo to pass it further but still the loop is possible if someone sign tx on cosmos and specify the memo with a loop to move assets back to cosmos. so now there is no validation to stop memo execution if the loop inside the memo. @blasrodri

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: memo is extracted to take root and execute pallet ibc extrinsic and pass the memo that inside 'next' field of current extracted memo. so we still need to serialise to pass new(unwraped memo) further.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: But once the transaction is returned, it won't contain any memo, right? Because we specified only one hop (cosmos->X->cosmos) and I don't see any reason to not allow it if it's not forbidden in the standard

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Why we cannot just pass the given `.next` memo? Without any (de)serialization

vmarkushin: If it's a different type, just add `T::MemoMessage: From<MemoData>` (or `TryFrom`) bound. I don't see any reason to use `serde_json` in this case

RustNinja:
> @RustNinja what's the status of this PR?

Yasin tested the functionality and we fix some issues. so i will change the code according to request changes and can merge it after.

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja:
```
mismatched types
expected enum `std::option::Option<<T as pallet::Config>::MemoMessage>`
   found enum `std::option::Option<&MemoData>`
```
I need to cast Memodata to T::MemoMessage type and it should be some json representation to be compatible on every chain to execute successfully  every hop.

contracts/pallet-ibc/src/ics20/memo.rs
RustNinja: Done.

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Let me know what to do with it. As i know cosmos chains does not reject any circuit memo.

contracts/pallet-ibc/Cargo.toml
RustNinja: Done

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Done

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Done

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Done

blasrodri:
Can you fix the compilation errors so that tests run?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
 use ibc::applications::transfer::error::Error as Ics20Error;
 
 pub trait HandleMemo<T: Config> {
-	fn execute_memo(packet_data: &PacketData, receiver: T::AccountId) -> Result<(), Ics20Error>;
+	fn execute_memo(packet: &Packet) -> Result<(), Ics20Error>;
```
blasrodri, line 566: why is the receiver no longer important?

RustNinja: You asked [here](https://github.com/ComposableFi/centauri/pull/359#discussion_r1245079395) to merge everything together and pass in execute_memo trait function.
So execute memo receives packet and then check memo and take receiver account to process memo.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Add `T::MemoMessage: From<MemoData>` and use `x.into()`

RustNinja: composable repo has his own MemoData struct for some XCVM integration.
so composable uses their own struct that equal to centauri MemoData..

contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -186,16 +184,16 @@ impl<T: Config + Send + Sync, S: Module + Clone + Default + PartialEq + Eq + Deb
...
-			serde_json::from_slice(packet.data.as_slice()).map_err(|e| {
-				Error::implementation_specific(format!("Failed to decode packet data {:?}", e))
-			})?;
-		let receiver = <T as Config>::AccountIdConversion::try_from(packet_data.receiver.clone())
-			.map_err(|_| {
-				Error::implementation_specific(format!("Failed to parse receiver account"))
-			})?
-			.into_account();
-		<T as Config>::HandleMemo::execute_memo(&packet_data, receiver).map_err(|e| {
+		// let packet_data: PacketData =
```
blasrodri, line 187: do we remove these?

RustNinja: yes. we removed and moved the logic about 
`let receiver = <T as Config>::AccountIdConversion::try_from(packet_data.receiver.clone())`
into execute_memo
to NOT call `serde_json::from_slice(packet.data.as_slice())` twice
here and inside of execute memo function

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+			})?;
+
+		//timestamp that current timestamp + 30 minutes
+		let _timestamp = <pallet_timestamp::Pallet<T> as frame_support::traits::UnixTime>::now()
+			.as_secs() + 1800;
+
+		let params = crate::TransferParams::<<T as frame_system::Config>::AccountId> {
+			to: transfer_ibc_account_to,
+			source_channel: channel_id,
+			// timeout: ibc_primitives::Timeout::Offset { timestamp: Some(timestamp), height: None
```
blasrodri, line 855: rm

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+						6,
+					);
+					Ics20Error::implementation_specific(
+						"failed to convert string to Config::MemoMessage".to_string(),
+					)
+				})?;
+			next_memo = Some(memo_result);
+		}
+
+		crate::Pallet::<T>::transfer(
```
blasrodri, line 887: I remember having some issues with middleware due to the packets not being removed from storage properly. Could this be a case here in case of failure @vmarkushin ?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+pub trait SubstrateMultihopXcmHandler {
+	type AccountId;
+
+	fn transfer_xcm(
+		from: Self::AccountId,
+		to: Self::AccountId,
+		para_id: Option<u32>,
+		amount: u128,
+		currency: u128,
+	) -> Option<()>;
```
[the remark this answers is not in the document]
RustNinja: pallet ibc does not know about Error type composable side.
So Result<(), ()> is equal Option<()>
I decided to have the Option<()> because in case if none centauri will emit event that error happened.

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: done.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Ok, what's the problem?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+					para_id: memo_forward.para_id,
+					amount,
+					asset_id: asset_id.into(),
+				});
+
+				return Ok(())
+			},
+		};
+
+		// let raw_bytes = memo_forward.receiver.as_bytes().to_vec();
```
vmarkushin, line 814: Remove?

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+
+use frame_system::RawOrigin;
+use scale_info::prelude::boxed::Box;
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct Forward {
+	pub receiver: String,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub port: Option<String>,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub channel: Option<String>,
```
vmarkushin, line 610: is it possible to use `ChannelId` and `PortId` from ibc-rs?

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: No, should be good

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+				Self::emit_memo_execution_failed_event(
+					receiver.clone(),
+					packet_data.memo.clone(),
+					4,
+				);
+				Ics20Error::implementation_specific("Failed to parse channel ID".to_string())
+			})?;
+
+		//timestamp that current timestamp + 30 minutes
+		let _timestamp = <pallet_timestamp::Pallet<T> as frame_support::traits::UnixTime>::now()
```
vmarkushin, line 849: The variable is not used anywhere, remove?

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Btw, if we use the same structure twice, it sounds like we can reuse one from the other side. 
And also, do we really need the `MemoMessage` type to be generic? Can't we just use one structure for every packet-forward?

RustNinja: This is because the same MemoData is using in XCVM and i think XCVM does not want to have a dependency from centauri and centauri are not able to depend on composable because of circuit dependency. 
So i decide to have two similar struct in composable and centauri once they are using only for deserialisation and serialisation.

RustNinja: And i would like to keep MessageType Generic,
because this is enum of memo forward, now it is XCM for substrate xcm hop and IBC for cosmos.
And who knows what other enum will be introduced by xcvm, etc.
So we basically can do it non generic but it will require the changes in both repos and several pallets each time when we will extend enum with a different variance.

contracts/pallet-ibc/src/ics20/memo.rs
```diff
@@ -186,16 +184,16 @@ impl<T: Config + Send + Sync, S: Module + Clone + Default + PartialEq + Eq + Deb
...
+		// 	serde_json::from_slice(packet.data.as_slice()).map_err(|e| {
+		// 		Error::implementation_specific(format!("Failed to decode packet data {:?}", e))
+		// 	})?;
+		// let receiver = <T as Config>::AccountIdConversion::try_from(packet_data.receiver.clone())
+		// 	.map_err(|_| {
+		// 		Error::implementation_specific(format!("Failed to parse receiver account"))
+		// 	})?
+		// 	.into_account();
```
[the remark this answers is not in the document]
RustNinja: Done.

contracts/pallet-ibc/src/ics20/mod.rs
```diff
@@ -561,11 +563,370 @@ pub fn full_ibc_denom(packet: &Packet, mut token: PrefixedCoin) -> String {
...
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub retries: Option<u64>,
+
+	/// since other parachain does not support ibc memo
+	/// there is only two option: send to parachain or send to relay-chain
+	// #[serde(skip_serializing_if = "Option::is_none")]
+	/// we do not need parrent id. if para id is none, it means send to relay-chain
+	// pub parent: Option<u32>,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub para_id: Option<u32>, //if para id is none, it means send to relay-chain
```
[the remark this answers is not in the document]
RustNinja: Done.

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Done.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: But isn't the MemoData initially comes from IBC? I mean [this](https://github.com/ComposableFi/composable/blob/32fa07505e3002796ee5a9564204c65409be5417/code/parachain/frame/composable-traits/src/xcm/memo.rs#L92) one. In this case, Composable should use the one from pallet-ibc. 

Anyway... Just using `From` trait should be enough for conversion between different types

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: Channel Id and PortId in memo is a simple String and not a struct.
I am not sure how that struct will be decode, encode to be compatible with memo format

contracts/pallet-ibc/src/ics20/mod.rs
RustNinja: In our case MemoData is our custom type that has a few different options for XCM forward hop and IBC forward hop.

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Ok. Then you should know that the channel is not always in the format `channel-xx`, it can even be called `my-channel` so parsing [here](https://github.com/ComposableFi/centauri/pull/359/files#diff-9c98c458f1edb9cac496e225691914b2ca600002781db177e670d624083bb592R827) may fail

vmarkushin: But if you need this specific format, you should be able to use ChannelId and PortId, because they de/serialize to correct strings

contracts/pallet-ibc/src/ics20/mod.rs
vmarkushin: Ok, but there shouldn't be any problems with that, no?

RustNinja: We tested. It works fine with a different option in memo (XCM, IBC hop) from cosmos chains though Picasso to composable.
Cosmos chain deserialise memo and did not fail and pass memo forward for futures hop.

blasrodri:
Can you check that if there’s errors the middleware handles them properly? Perhaps a unit test. We need to be sure that the sequences are removed if they have to be.

RustNinja:
> Can you check that if there’s errors the middleware handles them properly? Perhaps a unit test. We need to be sure that the sequences are removed if they have to be.

Done. added unit tests to check that do not mint or escrow any tokens and receiver balance has the expected amount of tokens.
validate that balance as expected and two extra event were emitted 
- memo execution started
- memo execution failed because incorrect memo(failed to deserialise) 

[on_deliver_ics20_recv_packet_incorrect_memo](https://github.com/ComposableFi/centauri/blob/packet-forward-middleware/contracts/pallet-ibc/src/tests.rs#L433C33-L566)

vmarkushin (approved):

blasrodri (approved):

PR #375, opened 2023-07-21
chore: changed docker image for builders

rjonczy:
Goal of this PR is to change docker image for ci jobs from:
`ci-linux:production` to `ci-unified:latest`

Issue: https://github.com/ComposableFi/centauri/issues/371

Using `:latest` tag: https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-unified#latest-tag as this is equivalent to `:production`, which we used before.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -15,7 +15,7 @@ jobs:
       group: lint-${{ github.ref }}
       cancel-in-progress: true
     container:
-      image: paritytech/ci-unified:latest
+      image: paritytech/ci-unified:bullseye-1.71.0
     runs-on:
       - self-hosted
       - x64-monster
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -15,7 +15,7 @@ jobs:
       group: lint-${{ github.ref }}
       cancel-in-progress: true
     container:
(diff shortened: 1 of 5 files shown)


.github/workflows/checks.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       group: lint-${{ github.ref }}
       cancel-in-progress: true
     container:
-      image: paritytech/ci-linux:production
+      image: paritytech/ci-unified:latest
```
vmarkushin, line 18: How about using a concrete version? Like `paritytech/ci-unified:bullseye-1.71.0-2023-05-23-v20230721`
This should prevent unexpected CI failures in case of breaking back-compatibility

blasrodri: I like that

blasrodri:
@rjonczy can you please close this one?

.github/workflows/checks.yml
rjonczy: @vmarkushin I agree. Actually I also think that using static tags is better than dynamic like: `:latest`

rjonczy:
> @rjonczy can you please close this one?

pushed fix commit

blasrodri (approved):

PR #377, opened 2023-07-24
Add a document for adding middleware to pallet-ibc

vmarkushin:
[Rendered](https://github.com/ComposableFi/centauri/blob/vmarkushin/middleware-docs/docs/middleware.md)

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -1,0 +1,96 @@
+## Adding a middleware on pallet-ibc
+
+### Overview
+
+The middleware is a module that wraps another module and adds some additional functionality to it. In the context of
+IBC, it means that the middleware is a module that is executed before/after the IBC module, and can access the packet
+data, take fees, send new packets etc. The middleware should be considered not a separate module, but rather a part of the module that 
+it's wrapping.
+
+To add a middleware for some IBC application (e.g., ics-20 - transfer app), one needs to create a new struct that 
+implements the `ibc::Module` trait and include it in the chain of other modules in runtime. Another important
+step is to make sure that the middleware [satisfies](#potential-issues) IBC packet flow to avoid unexpected behavior 
+of your app (for example, a possibility for double-spending).
+
+
+### Implementing `Module` trait and adding to runtime
+As an example, we'll take a look at the existing module for taking fees on transfers, `ics20_fee` (`contracts/pallet-ibc/src/ics20_fee`).
+First, we need to create a struct, that contains field `inner`, which is of type of the next module in the chain, so we can
+call it before/after our module execution is finished.
+```rust
+pub struct Ics20ServiceCharge<S: Module> {
+	inner: S,
+}
+```
+The `Module` contains various amount of callback methods that we can use, but probably the most important ones are 
+`on_recv_packet` and `on_acknowledgement_packet`. In our case, we want to take fee from the user when the packet is arrived
+and the transfer already happened, so the implementation will look something like this:
+```rust
+impl<S: Module> Module for Ics20ServiceCharge<S> {
+	fn on_recv_packet(
+		&self,
+		_ctx: &dyn ModuleCallbackContext,
+		output: &mut ModuleOutputBuilder,
+		packet: &mut Packet,
+		relayer: &Signer,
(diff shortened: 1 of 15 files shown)


docs/middleware.md
```diff
@@ -0,0 +1,109 @@
+## Adding a middleware on pallet-ibc
+
+### Overview
+
+The middleware is a module that wraps another module and adds some additional functionality to it. In the context of
+IBC, it means that the middleware is a module that is executed before/after the IBC module, and can access the packet
+data, take fees, send new packets etc. The middleware should be considered not a separate module, but rather a part of
+the module that
+it's wrapping.
```
blasrodri, line 9: so it's just one big module, right?

docs/middleware.md
```diff
@@ -0,0 +1,109 @@
...
+step is to make sure that the middleware [satisfies](#potential-issues) IBC packet flow to avoid unexpected behavior
+of your app (for example, a possibility for double-spending).
+
+### Implementing `Module` trait and adding to runtime
+
+As an example, we'll take a look at the existing module for taking fees on
+transfers, `ics20_fee` (`contracts/pallet-ibc/src/ics20_fee`).
+First, we need to create a struct, that contains an `inner` field, which is of type of the next module in the chain, so
+we can
+call it before/after our module execution is finished.
```
blasrodri, line 22: why before or after?

docs/middleware.md
vmarkushin: Yes, just with multiple layers, I would say

docs/middleware.md
```diff
@@ -0,0 +1,109 @@
...
+
+```rust
+pub struct Ics20ServiceCharge<S: Module> {
+    inner: S,
+}
+```
+
+The `Module` contains various amount of callback methods that we can use, but probably the most important ones are
+`on_recv_packet` and `on_acknowledgement_packet`. In our case, we want to take fee from the user when the packet arrives
```
blasrodri, line 31: how about timeout?

vmarkushin: Will add

docs/middleware.md
```diff
@@ -0,0 +1,109 @@
...
+    fn on_recv_packet(
+        &self,
+        _ctx: &dyn ModuleCallbackContext,
+        output: &mut ModuleOutputBuilder,
+        packet: &mut Packet,
+        relayer: &Signer,
+    ) -> Result<Acknowledgement, Error> {
+        let mut ctx = Context::<T>::default();
+        let ack = self.inner.on_recv_packet(&mut ctx, output, packet, relayer)?;
+        let _ = Self::process_fee(&mut ctx, packet, &ack).map_err(|e| {
```
blasrodri, line 45: actually the way to do this would be

`process_fee`, if it fails, we fail
if it succeeds, calls the inner
if the inner fails, we call `rollback_process_fees`

docs/middleware.md
```diff
@@ -0,0 +1,109 @@
...
+1. It's important to understand that the middleware is not a separate module, but rather a part of the module that it's
+   wrapping. This means that the middleware should not fail the execution after the execution of the `inner` module has
+   succeeded, because it may lead to unexpected behavior of the app. In our example, if the middleware fails (by
+   throwing an error), the packet won't be acknowledged, but the transfer will still happen. This means that even that
+   the tokens were transferred,
+   the whole packet is considered as failed and the tokens on the source chain will be returned back to the sender's
+   account. That's why we're ignoring the error from `process_fee` function, because we don't want the packet to fail
+   when the transfer already happened.
+
+2. Another important thing to note is that the middleware should not change the packet data, because it may lead to
```
blasrodri, line 107: we can enforce this, right? check that the hash of the packet hasn't changed at the beginning and end?

vmarkushin: The hash is already checked at this point. But the packet is passed via immutable reference, so to change it one would need to create a copy of the passed one, this change will be noticeable

docs/middleware.md
blasrodri: Can you may be added as a preferred alternative? Event if it's not merged

blasrodri (approved):

blasrodri (approved):

PR #382, opened 2023-07-31
oak-audit: fix issues 2, 12; Updated Grandpa light client with authority set history

vmarkushin:
The GRANDPA light-client logic and storage is updated to maintain a history of authorities set changes. This feature is necessary to support a more robust system, especially for evidence checks. In the previous version, the client could only store the current authorities, preventing it from correctly verifying whether misbehavior had taken place when a set change occurs.

The history of authorities is stored in the client state, with information of the block height, the timestamp of changes, the new set id and the new authorities.

This commit also enhances the misbehaviour detection functionality by checking every header in the submitted headers for evidence instead of only the first few unknown headers.

Additional fixes contain correcting the finality proof verification for client update and modifying relevant test logic accommodating the new client features.

Issue 2: "Impossible to report misbehaviors older than 500 blocks"
Instead of storing the last N headers, we now store all the headers for a fixed period of time (HEADER_ITEM_LIFETIME), which allows the fisherman to submit misbehavior even if validators reached the previous limit by submitting many blocks at once. Also, previous authority sets are now stored using the same method.

Also, misbehaviour validation was fixed: before the fix, it was only comparing the base blocks (they should have the same hash) and target block (hashes should be different), but it didn't check that these chains are actually diverging (one of them is a fork of another).


algorithms/grandpa/primitives/src/justification.rs
```diff
@@ -257,6 +257,33 @@ where
 	Ok(())
 }
 
+/// The function checks if the given chain is canonical:
```
blasrodri, line 260: great doc

contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -162,37 +158,32 @@ impl grandpa_client_primitives::HostFunctions for HostFunctionsManager {
 		pub_key.verify(&msg, sig)
 	}
 
-	fn insert_relay_header_hashes(new_hashes: &[<Self::Header as Header>::Hash]) {
+	fn insert_relay_header_hashes(now_ms: u64, new_hashes: &[<Self::Header as Header>::Hash]) {
```
blasrodri, line 161: why do you need to pass now_ms as an argument? Where does it come from?

hyperspace/parachain/src/provider.rs
```diff
@@ -808,4 +817,137 @@ where
...
+			.latest_height_and_timestamp()
+			.await
+			.map_err(|e| Error::Custom(e.to_string()))
+			.unwrap()
+			.0;
+		let latest_cp_client_state = counterparty
+			.query_client_state(latest_cp_height, client_id.clone())
+			.await
+			.map_err(|e| Error::Custom(e.to_string()))
+			.unwrap();
```
blasrodri, line 836: either map err or unwrap ser

PR #391, opened 2023-08-10
ci: fix hyperspace builds

blasrodri:
create /centauri/data dir while running as root user


.github/workflows/hyperspace-docker-image.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - fix-hyperspace-builds
```
rjonczy, line 7: remove before merging

rjonczy (dismissed):

PR #394, opened 2023-08-18
Refactor packet sequence handling in query methods

vmarkushin:
Replaced `HashSet` with `HashMap` for efficient and accurate packet sequence handling in `query_send_packets` and `query_received_packets` methods. The change optimizes the query process by ensuring that the latest block events are tracked correctly.


blasrodri (commented):
Why is this optimization not relevant for the Parachain?

vmarkushin:
> Why is this optimization not relevant for the Parachain?

In parachain we use a storage value instead of querying events

JafarAz (approved):

blasrodri (approved):

PR #400, opened 2023-08-30
setup parachain-launch

blasrodri:


dzmitry-lahoda:
added zombienet config, script and guide

PR #403, opened 2023-09-09
Fix decode_timestamp_extrinsic function to solve issue with incorrect timestamp decoded from extrinsic.

RustNinja:
investigation of test `test_continuous_update_of_grandpa_client`  that failed because of timestamp
Fix decode_timestamp_extrinsic function to solve issue with incorrect timestamp decoded from extrinsic.

### DIFF
--- a/light-clients/common/src/lib.rs
+++ b/light-clients/common/src/lib.rs
@@ -194,9 +194,9 @@ pub fn decode_timestamp_extrinsic(ext: &Vec<u8>) -> Result<u64, anyhow::Error> {
 	// https://github.com/paritytech/substrate/blob/d602397a0bbb24b5d627795b797259a44a5e29e9/primitives/trie/src/lib.rs#L99-L101
 	// Decoding from the [2..] because the timestamp inmherent has two extra bytes before the call
 	// that represents the call length and the extrinsic version.
-	let (_, _, timestamp): (u8, u8, Compact<u64>) = codec::Decode::decode(&mut &ext[2..])
+	let ret: Compact<u64> = codec::Decode::decode(&mut &ext[ext.len() - 7..])
 		.map_err(|err| anyhow!("Failed to decode extrinsic: {err}"))?;
-	Ok(timestamp.into())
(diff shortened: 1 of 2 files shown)


light-clients/common/src/lib.rs
```diff
@@ -194,9 +194,9 @@ pub fn decode_timestamp_extrinsic(ext: &Vec<u8>) -> Result<u64, anyhow::Error> {
 	// https://github.com/paritytech/substrate/blob/d602397a0bbb24b5d627795b797259a44a5e29e9/primitives/trie/src/lib.rs#L99-L101
 	// Decoding from the [2..] because the timestamp inmherent has two extra bytes before the call
 	// that represents the call length and the extrinsic version.
-	let (_, _, timestamp): (u8, u8, Compact<u64>) = codec::Decode::decode(&mut &ext[2..])
+	let ret: Compact<u64> = codec::Decode::decode(&mut &ext[ext.len() - 7..])
```
dzmitry-lahoda, line 197: just to note, i have sent to blas routine and way (via subxt type or runtime type trait combined official parsing method) to do the thing. i am for types and official code. i know people in this repo are not, as long as it works, awesome! just adding my 5 pica

RustNinja: @dzmitry-lahoda 
i was investigating the issue and just found where exactly it failed and use the simple way to fix to look what else effected or does not work at all.
now there is a new issue with `RootMismatch extrinsic proof, invalid proof: 0xb144…87f6`

so if you have a suggestion what is the best way to fix all the issues related to decoding please attach here.
really appreciate and thanks for your input.

dzmitry-lahoda: sure, yes  - what ever works . no worries:) that why i am not coding here :) 

i have dozen of work to do in CoDeFi area ;)

if any question about zombienet or process-compose - sure can help (but not with CI - it think to is to wild to manage and being sane).

in composable local runs = remote ci. so what runs locally runs remotely. so nice. no need to debug via jobs :)

PR #404, opened 2023-09-10
Fix timestamp issue

RustNinja:
fix `decode_timestamp_extrinsic` to decode timestamp directly into `Compact<u64>`


light-clients/common/src/lib.rs
```diff
@@ -194,9 +194,9 @@ pub fn decode_timestamp_extrinsic(ext: &Vec<u8>) -> Result<u64, anyhow::Error> {
 	// https://github.com/paritytech/substrate/blob/d602397a0bbb24b5d627795b797259a44a5e29e9/primitives/trie/src/lib.rs#L99-L101
 	// Decoding from the [2..] because the timestamp inmherent has two extra bytes before the call
```
nazar-pc, line 195: Comment should have been updated because now it doesnt make sense

PR #406, opened 2023-09-14
Increase nonce if tx failed. Not a fix. just a check

RustNinja:


RustNinja:

PR #409, opened 2023-09-15
Relayer skip tokens depends on config value: skip_tokens_list

RustNinja:
- introduced skip_tokens_list into config 
- use this list to check should relayer to skip the token

### DIFF
--- a/hyperspace/core/src/packets.rs
+++ b/hyperspace/core/src/packets.rs
@@ -356,7 +356,7 @@ pub async fn query_ready_and_timed_out_packets(
 						))
 					})?;
 
-					if list.iter().map(|x| decoded_dara.token.denom.base_denom.as_str() == x).any()) {
+					if list.iter().any(|skiped_denom| decoded_dara.token.denom.base_denom.as_str() == skiped_denom) {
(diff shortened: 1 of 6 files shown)


blasrodri:
Why do we need this?

vmarkushin (approved):

hyperspace/core/src/packets.rs
```diff
@@ -348,6 +349,20 @@ pub async fn query_ready_and_timed_out_packets(
...
+
+					let decoded_dara: PacketData = serde_json::from_str(&String::from_utf8_lossy(packet.data.as_ref())).map_err(|e| {
+						Error::Custom(format!(
+						"Failed to decode packet data for packet {:?}: {:?}",
+						packet, e
+						))
+					})?;
+
+					if list.iter().any(|skiped_denom| decoded_dara.token.denom.base_denom.as_str() == skiped_denom) {
+						log::warn!(target: "hyperspace", "Skipping packet as uosmo packet: {:?}", packet);
```
blasrodri, line 362: warn or info?

RustNinja:
> Why do we need this?

as i understood uosmo issue on the centauri side.

PR #417, opened 2023-09-20
Validate memo in `transfer` extrinsic

vmarkushin:
Added `ValidateMemo` trait


blasrodri (commented):
Can we add the supported one? So that the runtime can just imported an implementation? Where only osmosis IBC hooks work for now

PR #422, opened 2023-09-25
Fix client state substitution in ICS10-grandpa contract

vmarkushin:
Field `latest_relay_hash` is now also included in the state substitution process.

### DIFF
--- a/light-clients/ics10-grandpa-cw/src/contract.rs
+++ b/light-clients/ics10-grandpa-cw/src/contract.rs
@@ -269,16 +269,27 @@ fn process_message(
 
 			// Check that the substitute client state is valid:
 			// all fields should be the same as in the old state, except for the `relay_chain`,
-			// `para_id`, `latest_para_height`, `latest_relay_height`, `frozen_height`,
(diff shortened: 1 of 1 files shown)


light-clients/ics10-grandpa-cw/src/contract.rs
```diff
@@ -269,16 +269,27 @@ fn process_message(
...
+				latest_relay_height,
+				latest_relay_hash,
+				frozen_height,
+				latest_para_height,
+				para_id,
+				current_set_id,
+				current_authorities,
+				_phantom,
+			} = substitute_client_state.clone();
+			old_client_state.relay_chain = relay_chain;
```
blasrodri, line 285: Can we just add a function that does it? And in some way ensure that we are exhaustive. Perhaps just building an entire new one?

vmarkushin: This is a good way to make sure the we're exhaustive, because if we add new fields, the compilation will fail, because the fields are not mentioned here

blasrodri (approved):

PR #423, opened 2023-09-25
Local testing ibc parachain to parachain

RustNinja:
Subxt version 0.30.0 contained a huge major changes that produced the issue with nonce while sending transaction from relayer to parachain node.

method `next_nonce_account` was removed and a new method `nonce_account` works absolutely different way and because of it the error with nonce failed to submit tx time from time.

Solution: use version of subxt 0.29.0 (the prev tag pointed to 0.28.0) so applied the latest release before major changes in 0.30.0

Result: the is no error relate to `TooLowPriority { old: 13700, new: 513 }` and parachain to parachain test works as expected.


RustNinja:

PR #428, opened 2023-09-27
fix fmt and remove process-compose down in github ci

RustNinja:


RustNinja:

PR #439, opened 2023-10-24
relayer for Solana 

dhruvja:


hyperspace/solana/src/accounts.rs
```diff
@@ -0,0 +1,34 @@
+use anchor_client::anchor_lang::prelude::*;
+
+pub struct LocalDeliver {
```
blasrodri, line 3: Please add docs to structures

hyperspace/solana/src/instructions.rs
```diff
@@ -0,0 +1,41 @@
+use anchor_client::anchor_lang::prelude::*;
+use anchor_client::anchor_lang;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Deliver {
+  pub messages: Vec<AnyCheck>,
+}
+impl borsh::ser::BorshSerialize for Deliver
```
blasrodri, line 8: Isn't there a macro for this?

hyperspace/solana/src/instructions.rs
```diff
@@ -0,0 +1,41 @@
...
+  fn deserialize(
+      reader: &mut &[u8],
+  ) -> ::core::result::Result<Self, borsh::maybestd::io::Error> {
+      Ok(Self {
+          messages: borsh::BorshDeserialize::deserialize(reader)?,
+      })
+  }
+}
+impl anchor_lang::Discriminator for Deliver {
+  const DISCRIMINATOR: [u8; 8] = [250, 131, 222, 57, 211, 229, 209, 147];
```
blasrodri, line 33: Doc

hyperspace/solana/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+	) -> Result<
+		Pin<Box<dyn Stream<Item = <Self as IbcProvider>::FinalityEvent> + Send + Sync>>,
+		Error,
+	> {
+		todo!()
+	}
+
+	async fn submit(&self, messages: Vec<Any>) -> Result<Self::TransactionId, Error> {
+		let cluster = Cluster::from_str(&self.rpc_url).unwrap();
+		let authority = Rc::new(Keypair::from_bytes(&self.keybase.private_key).unwrap());
```
blasrodri, line 165: Why RC?

hyperspace/solana/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+				trie,
+				system_program::ID,
+			))
+			.args(instructions::Deliver { messages: all_messages })
+			.payer(authority.clone())
+			.signer(&*authority)
+			.send_with_spinner_and_config(RpcSendTransactionConfig {
+				skip_preflight: true,
+				..RpcSendTransactionConfig::default()
+			}).unwrap();
```
blasrodri, line 198: Better error handling?

hyperspace/solana/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+			authority.clone(),
+			CommitmentConfig { commitment: self.commitment_level },
+		);
+		let program = client.program(self.program_id).unwrap();
+
+		// Build, sign, and send program instruction
+		let storage_seeds = &[SOLANA_IBC_STORAGE_SEED];
+		let solana_ibc_storage = Pubkey::find_program_address(storage_seeds, &self.program_id).0;
+		let trie_seeds = &[TRIE_SEED];
+		let trie = Pubkey::find_program_address(trie_seeds, &self.program_id).0;
```
blasrodri, line 177: Why not get trie helper function or method?

blasrodri (commented):
Looking good. Ask a review from Vlad as well

hyperspace/solana/src/lib.rs
dhruvja: The docs used `Rc` but we can remove it since we only need immutable reference of the keypair.

hyperspace/solana/src/instructions.rs
dhruvja: Yes we can use the `AnchorSerialize` and `AnchorDeserialize` macro.

blasrodri:
Why are you planning to store data in the relayer? Ideally querying the node should suffice

dhruvja:
> Why are you planning to store data in the relayer? Ideally querying the node should suffice

I am not storing any data in the relayer. I am fetching the data from the accounts which are received from the rpc.

hyperspace/solana/src/lib.rs
```diff
@@ -555,7 +557,7 @@ impl IbcProvider for Client {
 			ibc::core::ics24_host::identifier::PortId,
 		)>,
 	) {
-		todo!()
```
blasrodri: Why do we need a lock?

dhruvja: because we are using a `Arc<Mutex<>>`

blasrodri: Why?

dhruvja: it was used in cosmos, so i thought it might be needed when we implement the methods further

blasrodri: I don't think we need to use a lock in Solana

dhruvja:
Since i am having issues with different IBC versions, I will try to fork it and upgrade the IBC for solana and cosmos relayer. Lets see how it goes.

mina86:
Upgrading is quite a big task though. What’s the main issue? It should be just a handful of types that we need from new ibc.

dhruvja:
I mean we just have to upgrade the `ibc_primitives` and the relayer for cosmos and solana.
I am done with `ibc_primitives` and `solana`

dhruvja:
Well looks like @mina86 was right and upgrading seems to be a big task. So i will prolly use the old version of ibc and the new one to fetch the data from the smart contract. Would also write some conversions to convert some types from new ibc to old one.

mina86:
Note that with 0.48 ibc-rs got split into multiple smaller crates, so you’re probably gonna be interested only in the crates which define types.

dhruvja:
> Note that with 0.48 ibc-rs got split into multiple smaller crates, so you’re probably gonna be interested only in the crates which define types.

gotcha

blasrodri (dismissed):

dzmitry-lahoda:
usually dockers are one liners to run

dhruvja:
Sure sure that could be done

PR #441, opened 2023-10-27
near: update borsh dependency from 0.9.3 to 0.10.3

mina86:


mina86:
@dhruvja, does that help us?

dhruvja:
Though it doesnt affect the borsh version in hyperspace, it might be a good idea to update it anyhow.

dhruvja (approved):

mina86:
@blasrodri, looks like I don’t have write permissions in the repository. Could you merge?

PR #442, opened 2023-10-27
Fix various clippy and compilation warnings

mina86:
In my previous commit I’ve introduced a new build warning.  Motivation
for this commit was to fix it and while I’m at it address some other
issues as well.  However, turns out there’s too many warnings I care
to deal with so this addresses only some of the warnings.


mina86:
@blasrodri

blasrodri (approved):

RustNinja (approved):

mina86:
@RustNinja, could you also merge? I don’t have write permissions in the repo.

PR #450, opened 2023-11-07
Ci optimise cloud costs

rjonczy:
Instead of using `e2-standard-32` in gcp, we can use Github runners.


blasrodri (commented):
What's the delta in time for the total CI to be finished with this change of machines?

rjonczy:
@blasrodri looks like build and publish time is very similar. What means that this jobs are not cpu or memory bound, but mostly is getting libraries/artifacts.

blasrodri (approved):

rjonczy:
Not merging yet. Some jobs are failing.

PR #456, opened 2023-11-20
create a docker build process for grandpa-cw contract

blasrodri:


blasrodri:
@rjonczy is this creating a new docker image?

.github/workflows/grandpa-cw-docker-image.yml
```diff
@@ -0,0 +1,74 @@
+name: "Build and publish Grandpa Docker image"
+
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - 'v*'
```
rjonczy, line 8: We need to first understand what is a purpose of this docker. Than we will be able to define when to build and publish this.

@blasrodri please do let me know when this image should be ready? At which stage of software development cycle?
- if there is any branch ? build and publish every commit?
- if there is any branch but with PR i.e. into master ?
- or maybe this is enough just to build and publish after merge into master branch ?
- or maybe this is enough to build and pubblich when building new release ? triggered by tag

.github/workflows/grandpa-cw-docker-image.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  
+  build-and-publish:
+
+    name: Build & push docker image
+    
+    runs-on:
+      - self-hosted
+      - x64-monster
```
rjonczy, line 21: By probably don't need to use self-hosted runner here. 

Lets try to use Github runners first:

```
runs-on: ubuntu-latest
```

scripts/grandpa-cw.Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENV PATH="/usr/local/protoc/bin:$PATH"
+
+RUN curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v21.9/protoc-21.9-linux-x86_64.zip && \
+	unzip  protoc-21.9-linux-x86_64.zip -d /usr/local/protoc && \
+	protoc --version
+
+
+# The .wasm file will live in /code/target/wasm32-unknown-unknown/release/ics10_grandpa_cw.wasm
+RUN RUSTFLAGS='-C link-arg=-s' cargo build -p ics10-grandpa-cw --target=wasm32-unknown-unknown --release --lib
```
rjonczy, line 15: no `ENTRYPOINT` ?

.github/workflows/grandpa-cw-docker-image.yml
```diff
@@ -0,0 +1,74 @@
...
+      uses: docker/setup-qemu-action@v3
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Extract metadata (tags, labels) for Docker
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.IMAGE_NAME }}
```
rjonczy, line 55: this cant' be the same like main image of `composable-ibc`.

Change:

```
env:
  IMAGE_NAME: ${{ github.repository }}
```

to something like:

```
env:
  IMAGE_NAME: composablefi/grandpa-cw
```

rjonczy (changes requested):

.github/workflows/grandpa-cw-docker-image.yml
blasrodri: Let's just build only with tags that contain the cosmos word. It's for the ibc-go developers right now

scripts/grandpa-cw.Dockerfile
blasrodri: The idea is just to have the contract built on the binary, so that people can mount it.

PR #461, opened 2023-12-19
regenerate subxt after RTU on composable to version 42

RustNinja:
cargo run -p codegen --bin codegen -- --path ./utils/subxt/generated/src/composable --relay-url wss://polkadot-rpc.dwellir.com --para-url wss://rpc.composable.finance


RustNinja:
did not affect relayer so do need to merge it. closed 
cc @vmarkushin

PR #488, opened 2024-03-29
Use types from upstream cf-guest

mina86:


mina86:
Depends on https://github.com/ComposableFi/emulated-light-client/pull/266.  Will update lock once that merges.

blasrodri (approved):

PR #490, opened 2024-04-02
Cleanup error enum

mina86:


mina86:
@blasrodri, can I haz LGTM? We need to get this merged to cw-solana an cw-solana merged for the audit.

PR #497, opened 2024-05-15
Weaken GRANDPA proof validation

vmarkushin:
`verify_with_voter_set` function is performing additional validation that go beyond the [standard implementation](https://raw.githubusercontent.com/w3f/consensus/master/pdf/grandpa.pdf) (section 2.3), possibly leading to proofs not being accepted while valid.

Instead, we need to accept all finality justifications where is_valid is true. The equivocations or other misbehaviors detected in the justifications must be handled too but not should not reject the proof.


blasrodri (approved):

kirillt:
Related discussion for reference: https://substrate.stackexchange.com/questions/8701/grandpa-finality-proofs-validation-and-equivocations

PR #504, opened 2024-06-26
adding new workflow

Ciejo:


vmarkushin (commented):
Does it support building images for both testnet and mainnet?Does it support building images for both testnet and mainnet?

Ciejo:
> Does it support building images for both testnet and mainnet?

No, it does not. Can you share with me the exact command to build each so I can write a workflow accordingly?

PR #510, opened 2024-08-05
ics-cf-solana light-client: header verification & client update; protobuf impl

vmarkushin:
Implemented initial version of the client header verification and state update, and Protobuf (de)serialisation.

The library is based on `cf-guest` crate and reuses some functionality from there. The verification of Shreds is performed by the functions taken from Solana's codebase. Some code was copy-pasted from the repo due to absence of no-std support, mentioning the original's file source path.

The current version of the code doesn't contain proper tests - these will be added when client is supported in hyperspace-solana.


light-clients/icsxx-cf-solana/src/client_def.rs
```diff
@@ -48,7 +49,51 @@ where
...
+					.iter()
+					.map(|shred| shred.sanitize().and_then(|_| shred.verify_with_root(&leader)))
+					.collect::<Result<Vec<_>, _>>()
+					.map_err(|e| {
+						Ics02ClientError::implementation_specific(alloc::format!(
+							"shred verification failed: {}",
+							e
+						))
+					})?;
+				Ok(())
```
mina86, line 90: Since you’re not doing anything with the shards collecting them into
Vec is waste of memory.

```suggestion
				header.shreds.iter().try_for_each(|shred| {
					shred.sanitize()?;
					shred.verify_with_root(&leader)
				}).map_err(|err| {
					Ics02ClientError::implementation_specific(alloc::format!(
						"shred verification failed: {err}",
					))
				})
```

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+	_phantom: PhantomData<PK>,
+}
+
+impl<PK> Header<PK> {
+	pub(crate) fn slot(&self) -> Slot {
+		assert!(!self.shreds.is_empty(), "Header must contain at least one shred");
+		self.shreds[0].slot()
+	}
+
+	pub fn hash(&self) -> Hash {
```
mina86, line 26:
```suggestion
	pub fn calculate_hash(&self) -> Hash {
```

So it’s clear that it’s an expensive function which calculates hash when called.

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+		let mut shreds = self.shreds.iter().collect::<Vec<_>>();
```
mina86, line 27: Unless I’m missing something, `shreds` is only used to make
`data_shreds`.  Based on that we probably should do filtering here so
that we don’t allocate unnecessary empty vector:

```suggestion
		let mut data_shreds = self.shreds
			.iter()
			.filter(|shred| shred.is_data())
			.collect::<Vec<_>>();
		data_shreds.sort_by_key(|s| s.index());
		data_shreds.dedup_by_key(|s| s.index());
```

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+							}))
+						},
+						None => return Err(BadMessage),
+					}
+				},
+				None => return Err(BadMessage),
+			};
+
+			shreds.push(shred);
+		}
```
mina86, line 261:
```suggestion
{
	{
		let shreds = msg.shreds.iter().map(|shred| match &proto_shred.message {
			Some(proto::shred::Message::ShredCode(proto_shred_code)) => {
				match &proto_shred_code.message {
					Some(proto::shred_code::Message::LegacyShredCode(
						proto_legacy_shred_code,
					)) => {
						// Convert LegacyShredCode
						let common_header: ShredCommonHeader = proto_legacy_shred_code
							.common_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let coding_header: CodingShredHeader = proto_legacy_shred_code
							.coding_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let payload = proto_legacy_shred_code.payload.clone();

						Shred::ShredCode(ShredCode::Legacy(legacy::ShredCode {
							common_header,
							coding_header,
							payload,
						}))
					},
					Some(proto::shred_code::Message::MerkleShredCode(
						proto_merkle_shred_code,
					)) => {
						// Convert MerkleShredCode
						let common_header: ShredCommonHeader = proto_merkle_shred_code
							.common_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let coding_header: CodingShredHeader = proto_merkle_shred_code
							.coding_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let payload = proto_merkle_shred_code.payload.clone();

						Shred::ShredCode(ShredCode::Merkle(merkle::ShredCode {
							common_header,
							coding_header,
							payload,
						}))
					},
					None => return Err(BadMessage),
				}
			},
			Some(proto::shred::Message::ShredData(proto_shred_data)) => {
				match &proto_shred_data.message {
					Some(proto::shred_data::Message::LegacyShredData(
						proto_legacy_shred_data,
					)) => {
						// Convert LegacyShredData
						let common_header: ShredCommonHeader = proto_legacy_shred_data
							.common_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let data_header: DataShredHeader = proto_legacy_shred_data
							.data_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let payload = proto_legacy_shred_data.payload.clone();

						Shred::ShredData(ShredData::Legacy(legacy::ShredData {
							common_header,
							data_header,
							payload,
						}))
					},
					Some(proto::shred_data::Message::MerkleShredData(
						proto_merkle_shred_data,
					)) => {
						// Convert MerkleShredData
						let common_header: ShredCommonHeader = proto_merkle_shred_data
							.common_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let data_header: DataShredHeader = proto_merkle_shred_data
							.data_header
							.as_ref()
							.ok_or(BadMessage)?
							.try_into()?;
						let payload = proto_merkle_shred_data.payload.clone();

						Shred::ShredData(ShredData::Merkle(merkle::ShredData {
							common_header,
							data_header,
							payload,
						}))
					},
					None => return Err(BadMessage),
				}
			},
			None => return Err(BadMessage),
		})
			.collect::<Result<Vec<_>, _>>()?;
```

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+
```
mina86, line 262: How about we enforce that shards i) are non-empty, ii) are sorted, iii) contain no duplicates and iv) belong to the same slot? This would simplify `hash` method and possibly other code? As far as I can tell there's no downside to requiring relayer to send us sorted unique data.

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+};
+use alloc::vec::Vec;
+use guestchain::PubKey;
+use proto_utils::BadMessage;
+use solana_sdk::{clock::Slot, hash::Hash, signature::Signature};
+use std::{collections::BTreeSet, convert::From, marker::PhantomData};
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct Header<PK> {
+	pub shreds: Vec<Shred>,
```
mina86, line 16: Could we live with this not being `pub` and having read-only accessor?

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,2 +1,397 @@
...
+	},
+};
+use alloc::vec::Vec;
+use guestchain::PubKey;
+use proto_utils::BadMessage;
+use solana_sdk::{clock::Slot, hash::Hash, signature::Signature};
+use std::{collections::BTreeSet, convert::From, marker::PhantomData};
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct Header<PK> {
```
mina86, line 15: If the header doesn’t depend on `PK` let just drop it?

light-clients/icsxx-cf-solana/src/proto/cf-solana.proto
```diff
@@ -0,0 +1,205 @@
...
+
+message Misbehaviour {
+  // First header.
+  Header header1 = 1;
+
+  // Second header.
+  //
+  // Note that the message may be partially filled.  If `header1` and
+  // `header2` have the same `genesis_hash` or `epoch` (which is extremely
+  // likely), those fields can be omitted in the second header.
```
mina86, line 201:
```suggestion
  // Second header.
```

light-clients/icsxx-cf-solana/src/solana/shred/legacy.rs
```diff
@@ -0,0 +1,263 @@
+//! File source: solana/ledger/src/shred/legacy.rs
```
mina86, line 1: I’m guessing we can get rid of legacy support?

mina86 (commented):
If I understand correctly, Header contains the entire Solana block (i.e. all shreds)? That can be rather big, no?

light-clients/icsxx-cf-solana/src/solana/shred/legacy.rs
vmarkushin: It's not being use in Solana anymore? Then I think it's crucial to make sure that the encoding is till correct, which seems a bit subtle to me... But I will add a TODO for it

light-clients/icsxx-cf-solana/src/client_def.rs
```diff
@@ -48,7 +49,51 @@ where
...
+					Ics02ClientError::implementation_specific(
+						"shreds have different slots".to_string(),
+					)
+				})?;
+
+				if slot <= client_state.latest_height().revision_height {
+					return Err(Ics02ClientError::implementation_specific(
+						"slot is not greater than latest height".to_string(),
+					));
+				}
```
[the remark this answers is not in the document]
vmarkushin: True. Also, some check for whether the header is in a trusted interval should be added I guess

vmarkushin:
> If I understand correctly, Header contains the entire Solana block (i.e. all shreds)? That can be rather big, no?

There is no requirement for that, but also no limit for how many shreds can be sent. I think it's reasonable to limit the amount of shreds only to be able to restore the last entry (to calculate the hash) + (potentially) some meta information, like validators change

light-clients/icsxx-cf-solana/src/client_def.rs
vmarkushin: Good call!

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: Good point. I'll create a special type for it

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: Will create getters

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: I thought maybe you'll use it for your part of work, but if you think it's not needed I'll remove it

mina86: PK would be used for places where signature is stored. This code uses Signature from solana_sdk so I don’t think there is a need for a generic.

light-clients/icsxx-cf-solana/src/solana/shred/legacy.rs
mina86: Honestly I don’t know, but that would be my assumption. It’s probably something to verify.

Cargo.toml
```diff
@@ -72,6 +72,8 @@ ibc-core-host = { git = "https://github.com/mina86/ibc-rs", rev = "2b50319f3460d
 ibc-core-host-types = { git = "https://github.com/mina86/ibc-rs", rev = "2b50319f3460d03e8781e5a834c4b56bdcfdc5bc", default-features = false }
 ibc-primitives = { git = "https://github.com/mina86/ibc-rs", rev = "2b50319f3460d03e8781e5a834c4b56bdcfdc5bc", default-features = false }
 ibc-testkit = { git = "https://github.com/mina86/ibc-rs", rev = "2b50319f3460d03e8781e5a834c4b56bdcfdc5bc", default-features = false }
+solana-sdk = { git = "https://github.com/ComposableFi/mantis-solana", rev = "d92c52575aaf430ca1f06bc2589f26b0fedde7bf" }
```
mina86, line 75: I’m a bit sceptical of using solana-sdk. The light client runs as a smart contract on Cosmos chain. Can we make this work with solana-program?

vmarkushin: It's just to reuse some definitions from there like VersionedTransaction. It supports no-std, so I guess we there won't be any issues deploying it as a cosmwasm contract

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: Removed the code because the checks are done in `PreCheckedShreds` now

light-clients/icsxx-cf-solana/src/client_def.rs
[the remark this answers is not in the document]
vmarkushin: Removed the checked and added a check for misbehaviour in this case

light-clients/icsxx-cf-solana/src/client_def.rs
```diff
@@ -48,7 +49,51 @@ where
...
+
+				let slot = shreds.first().unwrap().slot();
+
+				// All shreds' slots should be the same
+				shreds.iter().skip(1).all(|s| s.slot() == slot).then(|| ()).ok_or_else(|| {
+					Ics02ClientError::implementation_specific(
+						"shreds have different slots".to_string(),
+					)
+				})?;
```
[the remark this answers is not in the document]
vmarkushin: Removed the code because the checks are done in `PreCheckedShreds` now

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: The field has type `PreCheckedShreds` now, which is read-only

light-clients/icsxx-cf-solana/src/client_def.rs
```diff
@@ -130,12 +101,47 @@ where
...
-		_client_msg: Self::ClientMessage,
+		ctx: &Ctx,
+		client_id: ibc::core::ics24_host::identifier::ClientId,
+		client_state: Self::ClientState,
+		client_msg: Self::ClientMessage,
 	) -> Result<bool, Ics02ClientError> {
-		todo!("check_for_misbehaviour")
+		match client_msg {
+			ClientMessage::Header(header) => {
+				// The client can't be updated if no shreds were received
```
mina86, line 111: I’d split it into a new method.

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,68 +1,146 @@
...
 
-impl<PK: PubKey> From<Header<PK>> for proto::Header {
-	fn from(value: Header<PK>) -> Self {
+impl Deref for PreCheckedShreds {
+	type Target = [Shred];
+
+	fn deref(&self) -> &Self::Target {
+		&self.0
+	}
+}
```
mina86, line 85: This is something `derive_more::Deref` can do for you.

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,68 +1,146 @@
...
+				}
+			}
+
+			// Ensure the shreds don't contain duplicates.
+			if !shreds_set.insert(index) {
+				return Err(Error::ShredsContainDuplicates);
+			}
+
+			prev_index = Some(index);
+		}
```
mina86, line 131: Use `[T]::windows` instead.  This way you don’t have to index and deal
with prev_index being None.  I also don’t see need for shreds_set
since all indices are sorted.

```suggestion
		// TODO(mina86): use array_windows once it stabilises.
		for pair in shreds.iter().windows(2) {
			let prev = pair[0];
			let this = pair[1];

			if prev.slot() != this.slot() {
				return Err(Error::ShredsFromDifferentSlots);
			}

			if prev.index() > this.index() {
				return Err(Error::ShredsNotSorted);
			}

			if prev.index() == this.index() {
				return Err(Error::ShredsContainDuplicates);
			}
		}
```

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,68 +1,146 @@
...
+	}
+}
+
+/// An immutable array of shreds that have the following properties:
+/// 1. Is non-empty
+/// 2. Is sorted
+/// 3. Doesn't contain duplicates
+/// 4. All shreds have the same slot
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct PreCheckedShreds(Vec<Shred>);
```
mina86, line 69: I’d say `CheckedShreds` is a better name. `PreChecked` is ambiguous and might mean that those are shreds before the check, i.e. pre being checked.

light-clients/icsxx-cf-solana/src/header.rs
```diff
@@ -1,68 +1,146 @@
...
+
+			// Ensure the shreds don't contain duplicates.
+			if !shreds_set.insert(index) {
+				return Err(Error::ShredsContainDuplicates);
+			}
+
+			prev_index = Some(index);
+		}
+
+		Ok(Self(shreds))
```
mina86, line 133: Another thing is that we seem to only care about data shreds so maybe we should filter non-data shreds out?

```suggestion
		shreds.retain(Shred::is_data);
		Ok(Self(shreds))
```

light-clients/icsxx-cf-solana/src/client.rs
```diff
@@ -1,42 +1,71 @@
...
+// super::wrap!(cf_guest_upstream::ClientState as ClientState);
+// super::wrap!(impl proto for ClientState);
+
+impl ClientState {
+	pub(crate) fn timestamp_for_slot(&self, slot: Slot) -> Timestamp {
+		Timestamp::from_nanoseconds(
+			self.genesis_time.nanoseconds() + (slot as u64 * self.slot_duration.as_nanos() as u64),
+		)
+		.unwrap()
+	}
```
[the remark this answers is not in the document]
vmarkushin: You're right. Added a TODO

contracts/pallet-ibc/src/light_clients.rs
```diff
@@ -333,6 +337,7 @@ pub enum AnyClient {
 	Tendermint(ics07_tendermint::client_def::TendermintClient<HostFunctionsManager>),
 	Wasm(ics08_wasm::client_def::WasmClient<AnyClient, AnyClientState, AnyConsensusState>),
 	Guest(cf_guest::client_def::GuestClient<PubKey>),
+	CfSolana(icsxx_cf_solana::client_def::CfSolanaClient<PubKey>),
```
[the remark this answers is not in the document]
vmarkushin: Done

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: They're not fully checked (the signature wasn't checked at least), that's why the name is like so

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: Added a return error if some shreds are not data

light-clients/icsxx-cf-solana/src/header.rs
vmarkushin: What do you think about `PartiallyCheckedShreds`?

light-clients/icsxx-cf-solana/src/client.rs
```diff
@@ -1,42 +1,71 @@
...
+	pub current_leader: Pubkey,
 
-super::wrap!(cf_guest_upstream::ClientState<PK> as ClientState);
-super::wrap!(impl<PK> proto for ClientState);
+	/// Genesis timestamp
+	pub genesis_time: Timestamp,
 
-impl<PK: guestchain::PubKey> ClientState<PK> {
+	/// Chain's slot duration.
+	pub slot_duration: Duration,
```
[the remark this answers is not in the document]
vmarkushin: Applied

light-clients/icsxx-cf-solana/src/header.rs
mina86: Could also be `SortedShreds` I suppose?

PR #518, opened 2024-10-25
Update token supply from L1 to Rollup

godilov:
Mostly changed `rollup/lib.rs` and `rollup/client.rs`.

For `RollupClient` added RPC parameters to connect L1 - needed to request current token supply.

In `RollupClient.submit` added static `LAST_SUPPLY_UPDATE_TIME` (tried to add it in client, but `submit` takes immutable ref, so unable to update) and once a day update token supply.

### DIFF
--- a/hyperspace/rollup/src/lib.rs
+++ b/hyperspace/rollup/src/lib.rs
@@ -2050,7 +2050,7 @@ impl Chain for RollupClient {
 
 			let now = Instant::now();
 
-			if (now - *last).as_secs() > 24 * 60 * 60 {
+			if (now - *last).as_secs() > self.supply_update_time_in_sec {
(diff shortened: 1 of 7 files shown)


hyperspace/rollup/src/lib.rs
```diff
@@ -2044,6 +2049,45 @@ impl Chain for RollupClient {
 			// }
 		}
 
+		{
+			let mut last = LAST_SUPPLY_UPDATE_TIME.lock().await;
```
vmarkushin: Just add `last_supply_update_time: Arc<Mutex<Instant>>` to `RollupClient`

vmarkushin (commented):
The rest of the code LGTM

PR #520, opened 2024-10-25
hyperspace: pass witness account when calling solana-ibc

mina86:


vmarkushin (approved):

dhruvja:
@mina86 the witness feature is not required for solana. I think the name is a bit confusing but by solana i mean mainnet and rollup is mantis. Since they both are used in the same env we cannot have the witness feature. So thats why i pass the accounts as a vector in the rollup package. In case of solana, we dont need witness.

mina86:
kk, makes sense. I’ve been mostly concerned about the code not compiling and I guess didn’t look closely enough.

PR #522, opened 2024-10-28
Parse Mantis hook memo accounts and use them in a PacketRecv instruction (no SDK50)

vmarkushin:


hyperspace/solana/src/client.rs
```diff
@@ -704,29 +705,34 @@ deserialize consensus state"
...
+						escrow_account,
+						fee_collector: Some(self.get_fee_collector_key()),
+						receiver_token_account: receiver_address,
+						associated_token_program: Some(anchor_spl::associated_token::ID),
+						token_program: Some(anchor_spl::token::ID),
+					},
+					chunk_account,
+				);
+				let mut account_metas = accounts.to_account_metas(None);
+				if let Some(additional_accounts) = maybe_additional_accounts {
```
dhruvja: We only parse the memo if the base denom is `0x36dd1bfe89d409f869fabbe72c3cf72ea8b460f6`. Do u wanna check that as well?

vmarkushin: True

hyperspace/solana/src/client.rs
```diff
@@ -1103,6 +1109,72 @@ pub async fn get_accounts(
...
+	let accounts = strings
+		.drain(..count)
+		.map(Pubkey::from_str)
+		.map(Result::ok)
+		.collect::<Option<Vec<Pubkey>>>()?;
+	let account_metas = accounts
+		.into_iter()
+		.map(|pubkey| AccountMeta { pubkey, is_signer: false, is_writable: true })
+		.collect::<Vec<AccountMeta>>();
+	Some(account_metas)
```
mina86: The intermediary vector can be avoided quite easily.  Also, in this
instance you don’t want to use Vec::drain because that needs to resize
the source vector.

```suggestion
	strings
		.iter()
		.take(count)
		.map(|key| Pubkey::from_str(key).ok().map(|pubkey| AccountMeta {
			pubkey,
			is_signer: false,
			is_writable: true,
		})
		.collect()
```

(The `strings` vector could also be avoided but that’s a bit more code so probably not worth in the relayer).

hyperspace/solana/src/client.rs
```diff
@@ -1103,6 +1109,72 @@ pub async fn get_accounts(
...
+}
+
+#[test]
+fn test_parse_intent_memo_accounts() {
+	let memo = "2,Bt6A81KE6ZoWqG4SJa1gVCmcusSpmduJnPPpauRuNvAC,Hg3EhgX2USt8AzRNeX7Lh2uwTy3ij72qLeSofuCZK8Tc,137,{\"sender\":\"me\",\"receiver\":\"you\"}";
+	let accounts = parse_intent_memo_accounts(memo).unwrap();
+	assert_eq!(accounts.len(), 2);
+	assert!(accounts.iter().all(|account| account.is_writable));
```
mina86, line 1156: Should we also check is_signer?

hyperspace/solana/src/lib.rs
```diff
@@ -248,8 +248,8 @@ impl IbcProvider for SolanaClient {
 					let validator_idx = all_validators
 						.iter()
 						.position(|v| {
-							v.pubkey
-								== PubKey::from_bytes(&validator.to_bytes().as_slice()).unwrap()
+							v.pubkey ==
+								PubKey::from_bytes(&validator.to_bytes().as_slice()).unwrap()
```
mina86: I realise this is just formatting change so not really applicable to
this PR, but by the way, this probably could be simplified to

```suggestion
							v.pubkey.to_bytes() == validator.to_bytes()
```

mina86 (approved):
LGTM except for the denom check Dhruv mentioned.

hyperspace/solana/src/client.rs
vmarkushin: Fixed

hyperspace/solana/src/client.rs
vmarkushin: Fixed

hyperspace/solana/src/lib.rs
vmarkushin: Fixed

dhruvja (approved):

